PR #16, opened 2016-05-03
Lazy connect when booting from swagger generator

jannyHou:
Connect to strongloop/loopback#2242

When booting app from swagger generator, datasources are setup with ds.lazyConnect = true to avoid connection.

### DIFF
--- a/lib/cloudant.js
+++ b/lib/cloudant.js
@@ -14,7 +14,15 @@ var _ = require('lodash');
  */
 exports.initialize = function(ds, cb) {
   ds.connector = new Cloudant(ds.settings, ds);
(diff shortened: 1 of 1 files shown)


jannyHou:
@tonyffrench I am sorry if I didn't explain it clearly, so this pr is a sub-story of https://github.com/strongloop/loopback/issues/2242, which explains the original purpose and the solution.

I implement the feature in mongodb as an example(pr https://github.com/strongloop/loopback-connector-mongodb/pull/241), and now I am applying same change to other connectors.

And I looked through the test files but haven't decided which is the proper one to add test case for this feature. Any thought? 

Thanks.

tonyffrench:
LGTM

PR #27, opened 2016-09-01
Remove makefile & update eslint infrastructure

gunjpan:
- Using makefiles messes up coverage in CI.
npm test should be vanilla mocha [--options] (ideally mocha test/unit && test/integration).
 - make is not cross platform (ie. Windows)
 - CI has issues with coverage reports when we use make *

- Update eslint infrastructure to use `eslint-config-loopback` and remove `jscs`

Connect to https://github.com/strongloop-internal/scrum-loopback/issues/872

PR #29, opened 2016-09-07
Fix lint error and handle error

jannyHou:
- [x] Fix eslint failings.
- [x] Comment test cases to get CI pass until we have valid cloudant config.

### DIFF
--- a/test/maxrows.test.js
+++ b/test/maxrows.test.js
@@ -3,91 +3,94 @@
 // This file is licensed under the Artistic License 2.0.
 // License text available at https://opensource.org/licenses/Artistic-2.0
 
-'use strict';
+// Comment test cases to get CI pass,
+// will recover them when CI config done
 
-var should = require('should');
-var db, Thing;
+// 'use strict';
 
-describe('cloudant max rows', function() {
-  var Foo;
-  var N = 201;
-  before(function(done) {
-    require('./init.js');
-    db = getSchema();
-    Foo = db.define('Foo', {
-      bar: {type: Number, index: true},
-    });
-    Thing = db.define('Thing', {
-      title: Number,
-    });
-    Thing.belongsTo('foo', {model: Foo});
-    Foo.hasMany('things', {foreignKey: 'fooId'});
-    db.automigrate(done);
-  });
-  it('create two hundred and one', function(done) {
-    var foos = Array.apply(null, {length: N}).map(function(n, i) {
-      return {bar: i};
-    });
-    Foo.create(foos, function(err, entries) {
-      should.not.exist(err);
-      entries.should.have.lengthOf(N);
-      done();
-    });
-  });
-  it('find all two hundred and one', function(done) {
-    Foo.all(function(err, entries) {
-      if (err) return done(err);
-      entries.should.have.lengthOf(N);
-      var things = Array.apply(null, {length: N}).map(function(n, i) {
-        return {title: i, fooId: entries[i].id};
-      });
-      Thing.create(things, function(err, things) {
-        if (err) return done(err);
-        things.should.have.lengthOf(N);
-        done();
-      });
-    });
-  });
-  it('find all limt ten', function(done) {
-    Foo.all({limit: 10, order: 'bar'}, function(err, entries) {
-      if (err) return done(err);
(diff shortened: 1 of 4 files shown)


test/regexp.test.js
```diff
@@ -27,28 +28,31 @@ describe('cloudant regexp', function() {
...
       entries.should.have.lengthOf(1);
       entries[0].bar.should.equal('b');
       done();
     });
   });
   it('find all foos like b', function(done) {
     Foo.find({where: {bar: {like: 'b'}}}, function(err, entries) {
-      console.log (entries);
+      if (err) return done(err);
+      console.log(entries);
```
superkhau: Why did you move this down 2 lines?

test/regexp.test.js
```diff
@@ -27,28 +28,31 @@ describe('cloudant regexp', function() {
...
       entries[0].bar.should.equal('b');
       done();
     });
   });
   it('find all foos not like b', function(done) {
     Foo.find({where: {bar: {nlike: 'b'}}}, function(err, entries) {
-      console.log (entries);
-      entries.should.have.lengthOf(N-1);
+      if (err) done(err);
+      console.log(entries);
```
superkhau: ditto

test/maxrows.test.js
```diff
@@ -4,25 +4,27 @@
 // License text available at https://opensource.org/licenses/Artistic-2.0
 
 var should = require('should');
+var db, Thing;
```
superkhau: These weren't there before?

superkhau:
@jannyHou I'm suggesting you break this out into two commits. One for the ESLint specific fixes and one for the error fixes. I'm confused as to which lines of code are for what.

test/regexp.test.js
jannyHou: Eslint doesn't only complain syntax error but also errors like "error are supposed to be handled", this is a fix for error not handled. I am not moving this down 2 lines, I am adding a new line `if (err) return done(err)`

test/regexp.test.js
jannyHou: forgot a `return` here, will add it then.

test/maxrows.test.js
jannyHou: Nope, this is also reported by eslint, which means you could not use a variable without declaring it.

jannyHou:
@superkhau ok I will break them into different ones. But actually they are all errors reported by eslint, not fixing test cases.

And btw this pr is still in progress, mocha is not recognized in test files when running eslint, I am investigating how to get it work.

superkhau:
Sounds good, besides the missing return, LGTM.

test/regexp.test.js
superkhau: BTW, we should actually remove the console log line altogether. But that is outside the scope of this PR.

test/regexp.test.js
superkhau: Same as this console log.

test/regexp.test.js
```diff
@@ -27,28 +28,31 @@ describe('cloudant regexp', function() {
   });
   it('find all foos begining with b', function(done) {
     Foo.find({where: {bar: {regexp: '^b'}}}, function(err, entries) {
-      console.log (entries);
+      if (err) return done(err);
+      console.log(entries);
```
superkhau: Ditto for removing.

jannyHou:
@superkhau I just realized that this pr overlaps with the one from @gunjpan : 
https://github.com/strongloop/loopback-connector-cloudant/pull/27

Since he will be back after Sep.19 I am adding fix or commit upon his pr, closing this one now.

superkhau:
I think you should just fix the items I commented on already in this PR.

superkhau:
We can merge yours and he can rebase his later. Sept 11 is much later and these changes are 5 mins and I already reviewed it.

jannyHou:
@superkhau I commented the test cases to get CI pass. PTAL, thanks :)

superkhau:
LGTM, merge when all greenz. ;)

superkhau:
@jannyHou I think this needs backporting -- https://github.com/strongloop/loopback-connector-cloudant/issues/41

PR #30, opened 2016-09-19
Add translated files.

richardpringle:


jannyHou:
@superkhau @0candy @loay PTAL. Thanks.

PR #34, opened 2016-09-26
Add replaceOrCreate 

jannyHou:
Connect to strongloop-internal/scrum-loopback#991

- [x] Add function `replaceOrCreate`
  - Almost copy from `updateOrCreate`, which is actually doing replace not update.
  - At the end of the function, return data got from db.

- [x] Add function `replaceById`
  - Add an option `upsert` in function `create`, check id exist if `upsert === false`, default value is undefined.
  - `replaceById` will eventually call `replaceOrCreate` with option `upsert === false`

### DIFF
--- a/lib/cloudant.js
+++ b/lib/cloudant.js
@@ -140,19 +140,18 @@ Cloudant.prototype.fromDB = function(modelName, modelObject, doc) {
 };
 
 /**
- * Create a new model instance for the given data
+ * Insert a model instance
  *
  * @param {String} model The model name
  * @param {Object} data The model data
  * @param {Function} [cb] The cb function
  */
-Cloudant.prototype.create = function(model, data, options, cb) {
-  debug('Cloudant.prototype.create %j %j %j ', model, data, options);
+Cloudant.prototype._insert = function(model, data, cb) {
   var self = this;
   var idName = self.idName(model);
   var mo = self.selectModel(model);
   mo.db.insert(self.toDB(model, mo, data), function(err, result) {
-    debug('Cloudant.prototype.create insert %j %j', err, result);
+    debug('Cloudant.prototype.insert %j %j', err, result);
     if (err) {
       if (err.statusCode === 409) {
         err.message = err.message + ' (duplicate?)';
@@ -162,6 +161,18 @@ Cloudant.prototype.create = function(model, data, options, cb) {
     data[idName] = result.id;
     cb(null, result.id);
   });
+};
+
+/**
+ * Create a new model instance for the given data
+ *
+ * @param {String} model The model name
+ * @param {Object} data The model data
+ * @param {Function} [cb] The cb function
+ */
+Cloudant.prototype.create = function(model, data, options, cb) {
+  debug('Cloudant.prototype.create %j %j %j ', model, data, options);
+  this._insert(model, data, cb);
 };
 
 /**
@@ -576,6 +587,75 @@ Cloudant.prototype.ping = function(cb) {
     debug('Cloudant.prototype.ping results %j %j', err, result);
     if (err) cb('ping failed');
     else cb();
+  });
+};
+
+/**
+ * Replace if the model instance exists with the same id or create a
+ * new instance
+ *
+ * @param {String} model The model name
+ * @param {Object} data The model instance data
+ * @param {Function} [cb] The callback function
+ */
+Cloudant.prototype.replaceOrCreate = function replaceOrCreate(
+  model, data, options, cb) {
+  debug('Cloudant.prototype.replaceOrCreate %j %j', model, data);
+  var self = this;
+  var idName = self.idName(model);
+  var mo = self.selectModel(model);
+
+  var id = data[idName];
+  if (id) {
+    self.getCurrentRevision(model, id, function(err, rev) {
+      if (err) return cb(err);
+      if (rev) data._rev = rev;
+      self._insert(model, data, function(err) {
+        if (err) return cb(err);
+        mo.db.get(id, function(err, doc) {
+          if (err) return cb(err);
+          cb(err, self.fromDB(model, mo, doc), {isNewInstance: !rev});
+        });
+      });
+    });
+  } else {
+    self.create(model, data, options, function(err, id) {
+      if (err) return cb(err);
+      mo.db.get(id, function(err, doc) {
+        if (err) return cb(err);
+        cb(err, self.fromDB(model, mo, doc), {isNewInstance: true});
+      });
+    });
+  }
+};
+
+/**
+ * Replace properties for the model instance data
+ * @param {String} model The name of the model
+ * @param {*} id The instance id
+ * @param {Object} data The model data
+ * @param {Object} options The options object
+ * @param {Function} [cb] The callback function
+ */
+
+Cloudant.prototype.replaceById = function replaceById(
+  model, id, data, options, cb) {
+  debug('Cloudant.prototype.replaceById %j %j %j', model, id, data);
+  var self = this;
+  var mo = self.selectModel(model);
+  var idName = self.idName(model);
+  data[idName] = id;
+
+  self.getCurrentRevision(model, id, function(err, rev) {
+    if (err) return cb(err);
+    if (rev) data._rev = rev;
+    self._insert(model, data, function(err, id) {
+      if (err) return cb(err);
+      mo.db.get(id, function(err, doc) {
+        if (err) return cb(err);
+        cb(null, self.fromDB(model, mo, doc));
+      });
+    });
   });
 };
 
--- a/test/cloudant.test.js
+++ b/test/cloudant.test.js
@@ -1,0 +1,113 @@
+// Copyright IBM Corp. 2016. All Rights Reserved.
+// Node module: loopback-connector-cloudant
+// This file is licensed under the Artistic License 2.0.
+// License text available at https://opensource.org/licenses/Artistic-2.0
+
+// Comment test cases to get CI pass,
+// will recover them when CI config done
+
+'use strict';
+
+var should = require('should');
+var describe = require('./describe');
+var db, Product;
+
+describe('cloudant connector', function() {
+  before(function(done) {
+    db = getDataSource();
+
+    Product = db.define('Product', {
+      name: {type: String},
+      description: {type: String},
+      price: {type: Number},
+    }, {forceId: false});
+
+    Product.destroyAll(function(err) {
+      done();
+    });
+  });
+
+  describe('replaceOrCreate', function() {
+    it('should replace a model instance if the passing key already exists',
+      function(done) {
+        Product.create({
+          id: 1,
+          name: 'bread',
+          price: 100,
+          undefinedProperty: 'ShouldBeRemoved',
+        }, function(err, product) {
+          if (err) return done(err);
+          Product.replaceOrCreate({
+            id: product.id,
+            name: 'milk',
+          }, function(err, updatedProduct) {
+            if (err) return done(err);
+            verifyUpdatedData(updatedProduct);
+          });
+        });
+
+        function verifyUpdatedData(data) {
+          // Verify callback data
+          should.exist(data.id);
+          should.not.exist(data.price);
+          // Should remove extraneous properties not defined in the model
+          should.not.exist(data.undefinedProperty);
+          data.name.should.be.equal('milk');
+
+          // Verify DB data
+          verifyDBData(data.id);
+        };
+
+        function verifyDBData(id) {
+          Product.findById(id, function(err, data) {
+            if (err) return done(err);
+            should.not.exist(data.price);
+            should.not.exist(data.undefinedProperty);
+            data.name.should.be.equal('milk');
+            done();
+          });
+        };
+      });
+  });
+
+  describe('replaceById', function() {
+    it('should replace the model instance if the provided key already exists',
+      function(done) {
+        Product.create({
+          id: 2,
+          name: 'bread',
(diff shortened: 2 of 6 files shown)


jannyHou:
@Amir-61 @bajtos PTAL. Thanks :)  Details see comment above.

lib/cloudant.js
```diff
@@ -580,6 +590,59 @@ Cloudant.prototype.ping = function(cb) {
...
+ * @param {Object} data The model data
+ * @param {Object} options The options object
+ * @param {Function} [cb] The callback function
+ */
+
+Cloudant.prototype.replaceById = function replaceById(
+  model, id, data, options, cb) {
+  data.id = id,
+  options.upsert = false;
+  this.replaceOrCreate(model, data, options, cb);
```
Amir-61: I'm not sure if this is the valid logic...
Shouldn't `replaceById` has its own implementation instead of a wrapper method around `replaceOrCreate`? In other words I do not think if `replaceById`should create the instance if it does not exist (that would be the logic for `replaceOrCreate`)

test/cloudant.test.js
```diff
@@ -0,0 +1,114 @@
...
+
+// Comment test cases to get CI pass,
+// will recover them when CI config done
+
+'use strict';
+
+var should = require('should');
+var db, Product;
+
+describe.skip('cloudant connector', function() {
```
Amir-61: Is there any reason why you are skipping the tests?

lib/cloudant.js
```diff
@@ -580,6 +590,59 @@ Cloudant.prototype.ping = function(cb) {
...
+    });
+  } else {
+    self.create(model, data, options, function(err, id) {
+      if (err) return cb(err);
+      self.findById(model, data.id, {}, function(err, result) {
+        if (err) return cb(err);
+        cb(err, result, {isNewInstance: true});
+      });
+    });
+  }
```
Amir-61: I'm afraid I'm not familiar with cloundantDB; but AFICT this code is prone to race condition; however, I see they have used the same pattern for other methods as well... So you may land it this way and fix these race conditions with future PRs for `replaceOrCreate` and other methods where there are race conditions

test/cloudant.test.js
```diff
@@ -0,0 +1,114 @@
...
+  describe('replaceOrCreate', function() {
+    it('should replace a model instance if the passing key already exists',
+      function(done) {
+        Product.create({
+          id: 1,
+          name: 'bread',
+          price: 100,
+          undefinedProperty: 'ShouldBeRemoved',
+        }, function(err, product) {
+          console.log('create====== ', err);
```
Amir-61: What is this console.log? Shouldn't you remove it?

test/cloudant.test.js
```diff
@@ -0,0 +1,114 @@
...
+          if (err) return done(err);
+          Product.replaceOrCreate({
+            id: product.id,
+            name: 'milk',
+          }, function(err, updatedProduct) {
+          	 console.log('replace====== ', err);
```
Amir-61: ditoo

test/cloudant.test.js
```diff
@@ -0,0 +1,114 @@
...
+          if (err) return done(err);
+          Product.replaceById(product.id, {name: 'apple'},
+            function(err, updatedProduct) {
+              if (err) return done(err);
+              verifyUpdatedData(updatedProduct);
+            });
+        });
+
+        function verifyUpdatedData(data) {
+	      // Verify callback data
```
Amir-61: Indentation is off; please fix it

lib/cloudant.js
Amir-61: >In other words <del>I do not think if replaceByIdshould create the instance if it does not exist</del> (that would be the logic for replaceOrCreate

Oh I just noticed you are using `options.upsert = false;`

lib/cloudant.js
```diff
@@ -580,6 +590,59 @@ Cloudant.prototype.ping = function(cb) {
...
+ * @param {String} model The name of the model
+ * @param {*} id The instance id
+ * @param {Object} data The model data
+ * @param {Object} options The options object
+ * @param {Function} [cb] The callback function
+ */
+
+Cloudant.prototype.replaceById = function replaceById(
+  model, id, data, options, cb) {
+  data.id = id,
```
jannyHou: Thanks will try `self.idName()`

test/cloudant.test.js
jannyHou: All tests are skipped atm. CI is not setup yet.

lib/cloudant.js
jannyHou: I create a bug issue to track race condition fix: https://github.com/strongloop/loopback-connector-cloudant/issues/38

Will provide more details when someone get chance to work on it.

lib/cloudant.js
```diff
@@ -151,17 +151,27 @@ Cloudant.prototype.create = function(model, data, options, cb) {
   var self = this;
   var idName = self.idName(model);
   var mo = self.selectModel(model);
-  mo.db.insert(self.toDB(model, mo, data), function(err, result) {
-    debug('Cloudant.prototype.create insert %j %j', err, result);
-    if (err) {
-      if (err.statusCode === 409) {
-        err.message = err.message + ' (duplicate?)';
+  if (options.upsert === false) {
+    self.findById(model, data.id, {}, function(err) {
```
bajtos: I find this confusing. Why is `create` method querying the database? What if `data.id` is not set, because it's auto-generated by the database? 

The property name `id` should not be hard-coded.

`Cloudant.prototype.create` is a public method used by DAO method in juggler. I don't think it's a good idea to add `options.support` to a method like this, because now users can do something like this:

```
Product.create({name: 'Pen' }, { upsert: false }, cb);
```

(No other connectors support `options.upsert` in `create` to my best knowledge.)

lib/cloudant.js
bajtos: This looks suspicious to me too. IIUC, `replaceOrCreate` will call `create` under the hood, which will in turn call `findById` and `insertData`. That is defying one of the goals of `replaceId`, which is to get rid of the extra `findById` query.

I would much more prefer to see `replaceById` calling directly `insertData` (or similar).

test/cloudant.test.js
bajtos: Can we use the approach where the tests are skipped only when running on CI, so that developers running the `npm test` locally can still see test results?

See https://github.com/strongloop/loopback-connector-informix/pull/9/files

jannyHou: > Can we use the approach where the tests are skipped only when running on CI, so that developers running the npm test locally can still see test results?

Sounds Good.

lib/cloudant.js
jannyHou: @bajtos I get the idea from mongodb:

`MongoDB.prototype.replaceById` uses an option `upsert` to decide check id existed or not, see:
https://github.com/strongloop/loopback-connector-mongodb/blob/master/lib/mongodb.js#L962

But true, since Cloudant.prototype.create is a public method used by DAO it's dangers to expose that option...

jannyHou:
Update:

- [ ] Added a new private function `Cloudant.prototype._insert()`
- [ ] `create()` and `replacById` both call `_insert()` and `replaceOrCreate()` calls `create()`

@bajtos PTAL again. Thanks!

cc @Amir-61

lib/cloudant.js
```diff
@@ -140,19 +140,18 @@ Cloudant.prototype.fromDB = function(modelName, modelObject, doc) {
...
  */
-Cloudant.prototype.create = function(model, data, options, cb) {
-  debug('Cloudant.prototype.create %j %j %j ', model, data, options);
+Cloudant.prototype._insert = function(model, data, cb) {
   var self = this;
   var idName = self.idName(model);
   var mo = self.selectModel(model);
   mo.db.insert(self.toDB(model, mo, data), function(err, result) {
-    debug('Cloudant.prototype.create insert %j %j', err, result);
+    debug('Cloudant.prototype.insert %j %j', err, result);
```
Amir-61: Nitpick:

I would probably make them more explicit like:
```
debug('Cloudant.prototype._insert err:%j, result: %j', err, result);
```

Please feel free to ignore it

lib/cloudant.js
```diff
@@ -637,9 +641,22 @@ Cloudant.prototype.replaceOrCreate = function replaceOrCreate(
...
+  debug('Cloudant.prototype.replaceById %j %j %j', model, id, data);
+  var self = this;
+  var mo = self.selectModel(model);
+  var idName = self.idName(model);
+  data[idName] = id;
+
+  self.getCurrentRevision(model, id, function(err, rev) {
+    if (err) return cb(err);
+    if (rev) data._rev = rev;
+    self._insert(model, data, function(err, id) {
```
Amir-61: Shouldn't you check `err` here as well?

Amir-61: I mean:
```
if (err) return cb(err);
```

lib/cloudant.js
```diff
@@ -165,6 +164,19 @@ Cloudant.prototype.create = function(model, data, options, cb) {
...
+ * Create a new model instance for the given data
+ *
+ * @param {String} model The model name
+ * @param {Object} data The model data
+ * @param {Function} [cb] The cb function
+ */
+Cloudant.prototype.create = function(model, data, options, cb) {
+  debug('Cloudant.prototype.create %j %j %j ', model, data, options);
+  var self = this;
+  self._insert(model, data, cb);
```
Amir-61: Can we change:
```
  var self = this;
  self._insert(model, data, cb);
```
to:
```
  this._insert(model, data, cb);
```

lib/cloudant.js
jannyHou: Good catch. added

lib/cloudant.js
jannyHou: I will submit a pr to reword debug msg to provide more details. But it's out of the scope of this pr lol. Thanks.

lib/cloudant.js
jannyHou: The original implementation ignores options while I think in the future we should take care of it, so maybe at that time we will need var self = this.

But thanks now i will change to the simple one.

lib/cloudant.js
```diff
@@ -580,6 +591,75 @@ Cloudant.prototype.ping = function(cb) {
...
+  var self = this;
+  var idName = self.idName(model);
+  var mo = self.selectModel(model);
+
+  var id = data[idName];
+  if (id) {
+    self.getCurrentRevision(model, id, function(err, rev) {
+      if (err) return cb(err);
+      if (rev) data._rev = rev;
+      self.create(model, data, options, function(err) {
```
bajtos: Should we call `this._insert` instead?

test/cloudant.test.js
```diff
@@ -0,0 +1,113 @@
...
+          if (err) return done(err);
+          Product.replaceById(product.id, {name: 'apple'},
+            function(err, updatedProduct) {
+              if (err) return done(err);
+              verifyUpdatedData(updatedProduct);
+            });
+        });
+
+        function verifyUpdatedData(data) {
+        // Verify callback data
```
bajtos: Nitpick: the indentation of this comment is wrong, same applies to the two comments below.

test/imported.test.js
```diff
@@ -6,14 +6,15 @@
...
+    require('./init.js');
+  });
 
-//   require ('loopback-datasource-juggler/test/include.test.js');
-//   require ('loopback-datasource-juggler/test/crud-with-options.test.js');
-//   require ('loopback-datasource-juggler/test/common.batch.js');
-// });
+  require ('loopback-datasource-juggler/test/include.test.js');
+  require ('loopback-datasource-juggler/test/crud-with-options.test.js');
+  require ('loopback-datasource-juggler/test/common.batch.js');
```
bajtos: It's outside of scope of this pull request, but the "imported features" suite should include tests for operation hooks too. See e.g. https://github.com/strongloop/loopback-connector-mysql/blob/18cffb98ab5a90b8e728019de2b9b050fdc8a383/test/persistence-hooks.test.js

lib/cloudant.js
jannyHou: Now they are almost same, but true `replaceOrCreate` should keep identity to `this._insert()` instead of `this.create()`. Will change it.

jannyHou:
Feedbacks applied. @Amir-61  PTAL again :) Thanks.

bajtos:
@jannyHou Please rebase the branch on top of the latest master.

test/regexp.test.js
```diff
@@ -6,65 +6,66 @@
...
+  var Foo;
+  var N = 10;
+  before(function(done) {
+    db = getSchema();
+    Foo = db.define('Foo', {
+      bar: {type: String, index: true},
+    });
+    db.automigrate(done);
+  });
+  it('create some foo', function(done) {
```
Amir-61: * Missing `s`; it should be:
`creates some foo`
* Also I would use a more meaningful name than `foo`and `bar`

This comment applies for all your tests

Please feel free to ignore it.

Amir-61:
Please squash all commits as one single commit and rebase it on top of master and it seems you need to deal with some merge conflicts. LGTM.

test/regexp.test.js
jannyHou: Sorry this is not committed me...I created an issue to improve the debug and test case message: https://github.com/strongloop/loopback-connector-cloudant/issues/42

I will track all refactor related things there.
Won't make the change in this pr though, out of the scope, it's not a big job but if we keep focusing on similar things then the pr will do endless change lol.

PR #54, opened 2016-11-08
replaceOrCreate should create new instance if id provided

jannyHou:
Connect to https://github.com/strongloop-internal/scrum-loopback#991

Scenario:
Without fix in this pr
```js
MyModel.replaceOrCreate({
  id: 'myid', // id provided 
  name: 'myname',
}, function(err, result){...});
```
will fail if the instance doesn't exist in database.
The current code doesn't check instance exist before do the [`getRevision()`](https://github.com/strongloop/loopback-connector-cloudant/blob/master/lib/cloudant.js#L610) , which cannot run against an nonexistent instance.

Details see: https://github.com/strongloop-internal/scrum-loopback/issues/991#issuecomment-256077756

### DIFF
--- a/lib/cloudant.js
+++ b/lib/cloudant.js
@@ -604,29 +604,39 @@ Cloudant.prototype.replaceOrCreate = function replaceOrCreate(
   var self = this;
   var idName = self.idName(model);
   var mo = self.selectModel(model);
-
   var id = data[idName];
-  if (id) {
-    self.getCurrentRevision(model, id, function(err, rev) {
-      if (err) return cb(err);
-      if (rev) data._rev = rev;
-      self._insert(model, data, function(err) {
+
+  // Check if data exist first to avoid
+  // running getCurrentRevision() again an nonexistent instance
+  var where = {}; where[idName] = id;
+  var dataExist;
+
+  self.count(model, where, {}, function(err, cnt) {
+    if (err) return cb(err);
+    dataExist = cnt === 1 ? true : false;
+
+    if (id && dataExist === true) {
+      self.getCurrentRevision(model, id, function(err, rev) {
+        if (err) return cb(err);
+        if (rev) data._rev = rev;
+        self._insert(model, data, function(err) {
+          if (err) return cb(err);
+          mo.db.get(id, function(err, doc) {
+            if (err) return cb(err);
+            cb(err, self.fromDB(model, mo, doc), {isNewInstance: !rev});
+          });
+        });
+      });
+    } else {
+      self.create(model, data, options, function(err, id) {
         if (err) return cb(err);
         mo.db.get(id, function(err, doc) {
           if (err) return cb(err);
-          cb(err, self.fromDB(model, mo, doc), {isNewInstance: !rev});
+          cb(err, self.fromDB(model, mo, doc), {isNewInstance: true});
         });
       });
-    });
-  } else {
-    self.create(model, data, options, function(err, id) {
-      if (err) return cb(err);
-      mo.db.get(id, function(err, doc) {
-        if (err) return cb(err);
-        cb(err, self.fromDB(model, mo, doc), {isNewInstance: true});
-      });
-    });
-  }
+    };
+  });
 };
 
 /**
--- a/test/cloudant.test.js
+++ b/test/cloudant.test.js
@@ -68,6 +68,36 @@ describe('cloudant connector', function() {
           });
         };
       });
+    it('should create new instance if id provided', function(done) {
+      Product.replaceOrCreate({
+        id: 2,
+        name: 'newInstance',
+        price: 100,
+      }, function(err, product) {
+        if (err) return done(err);
+        verifyCreatedData(product);
+      });
+
+      function verifyCreatedData(data) {
+        // Verify callback data
+        data.id.should.be.equal('2');
+        data.name.should.be.equal('newInstance');
+        data.price.should.be.equal(100);
+
+        // Verify DB data
+        verifyDBData(data.id);
+      };
+
+      function verifyDBData(id) {
+        Product.findById(id, function(err, data) {
+          if (err) return done(err);
+          data.id.should.be.equal('2');
+          data.name.should.be.equal('newInstance');
+          data.price.should.be.equal(100);
+          done();
+        });
+      };
+    });
   });
 
   describe('replaceById', function() {


jannyHou:
The change comparison looks messy..
Let me explain a little bit what I changed in this pr:
I just wrapped `if(id){...}else{...}` with `self.count()`, so now the logic turns to:

```js
self.count({condition_to_check_instance_exist}, function(err, cnt) {
  dataExist = cnt === 1 ? true : false;
  if (id && dataExist) {
    // Do Replace, make sure getRevision won't run on an nonexistent instance
  } else {
    // Do Create
  }
})

```

test/cloudant.test.js
```diff
@@ -68,6 +68,36 @@ describe('cloudant connector', function() {
...
+      Product.replaceOrCreate({
+        id: 2,
+        name: 'newInstance',
+        price: 100,
+      }, function(err, product) {
+        if (err) return done(err);
+        verifyCreatedData(product);
+      });
+
+      function verifyCreatedData(data) {
```
Amir-61: I would chose a better function name; how about `verifyReturnedCallbackData`?

test/cloudant.test.js
```diff
@@ -68,6 +68,36 @@ describe('cloudant connector', function() {
...
+        // Verify callback data
+        data.id.should.be.equal('2');
+        data.name.should.be.equal('newInstance');
+        data.price.should.be.equal(100);
+
+        // Verify DB data
+        verifyDBData(data.id);
```
Amir-61: I believe the method name is/should be descriptive enough so the comment is unnecessary in this case...

test/cloudant.test.js
```diff
@@ -68,6 +68,36 @@ describe('cloudant connector', function() {
...
+      };
+
+      function verifyDBData(id) {
```
Amir-61: Nitpick: method name is good but I may still choose `verifyPersistedData`... I leave it to you...Please feel free to ignore this suggestion...

test/cloudant.test.js
```diff
@@ -68,6 +68,36 @@ describe('cloudant connector', function() {
           });
         };
       });
+    it('should create new instance if id provided', function(done) {
```
Amir-61: Feel free to ignore since in this file other tests use the same pattern; however, in general we use imperative mood, we do not start test names with should. Please see: https://github.com/strongloop/loopback.io/blob/gh-pages/pages/en/contrib/style-guide.md#test-naming

lib/cloudant.js
```diff
@@ -604,29 +604,39 @@ Cloudant.prototype.replaceOrCreate = function replaceOrCreate(
...
+  // Check if data exist first to avoid
+  // running getCurrentRevision() again an nonexistent instance
+  var where = {}; where[idName] = id;
+  var dataExist;
+
+  self.count(model, where, {}, function(err, cnt) {
+    if (err) return cb(err);
+    dataExist = cnt === 1 ? true : false;
+
+    if (id && dataExist === true) {
```
Amir-61: How about:
```
if (id && dataExist) {
   ...
}
```

lib/cloudant.js
```diff
@@ -604,29 +604,39 @@ Cloudant.prototype.replaceOrCreate = function replaceOrCreate(
...
+      self.getCurrentRevision(model, id, function(err, rev) {
         if (err) return cb(err);
-        mo.db.get(id, function(err, doc) {
+        if (rev) data._rev = rev;
+        self._insert(model, data, function(err) {
           if (err) return cb(err);
-          cb(err, self.fromDB(model, mo, doc), {isNewInstance: !rev});
+          mo.db.get(id, function(err, doc) {
+            if (err) return cb(err);
+            cb(err, self.fromDB(model, mo, doc), {isNewInstance: !rev});
```
Amir-61: I think in this branch `rev` always exists and `!rev` would always be `false`, right? In this case wouldn't it be more clear if you pass `{isNewInstance: false}` instead? Is my understanding right here? Correct me if I'm wrong.

lib/cloudant.js
```diff
@@ -604,29 +604,39 @@ Cloudant.prototype.replaceOrCreate = function replaceOrCreate(
...
-
   var id = data[idName];
-  if (id) {
-    self.getCurrentRevision(model, id, function(err, rev) {
-      if (err) return cb(err);
-      if (rev) data._rev = rev;
-      self._insert(model, data, function(err) {
+
+  // Check if data exist first to avoid
+  // running getCurrentRevision() again an nonexistent instance
```
Amir-61: Check if data exist**s** to avoid running 'getCurrentRevision()' against new data.

lib/cloudant.js
```diff
@@ -604,29 +604,39 @@ Cloudant.prototype.replaceOrCreate = function replaceOrCreate(
...
+  var where = {}; where[idName] = id;
+  var dataExist;
```
Amir-61:
```
var where = {};
where[idName] = id;
var dataExist;
```

lib/cloudant.js
```diff
@@ -604,29 +604,39 @@ Cloudant.prototype.replaceOrCreate = function replaceOrCreate(
...
-        cb(err, self.fromDB(model, mo, doc), {isNewInstance: true});
+        mo.db.get(id, function(err, doc) {
+          if (err) return cb(err);
+          cb(err, self.fromDB(model, mo, doc), {isNewInstance: true});
+        });
       });
-    });
-  }
+    };
+  });
```
Amir-61: I'm not that much familiar with cloudant; is this the only way to do it in cloudant? Isn't there any atomic method to use? I trust your judgment in this regard...

test/cloudant.test.js
Amir-61: You may use 'Primary key' instead of `id`; also you could add a test when the Primary key is not called `id` as well...

bajtos:
If I understand this change correctly, then each `replaceOrCreate` operation is making three database calls now: `count`, `getRevision`, replaceById or create. That creates a long window where another request may come in and interfere with our operation in progress :( Is this implementation any better than using the naive algorithm from juggler (`findById` + replaceById or create)? If my understanding is correct, then neither of these implementations is truly atomic.

When I step back a bit and look at the problem from a high level perspective, then I think it should be acceptable to require callers of patch/replace operations to provide `_rev` in the data, as that's the way how CouchDB/Cloudant allows clients to detect and resolve conflicts. With this requirement in place, the implementation in this connector should be much simpler - no need to call `getRevision`, just call `_insert`  directly.

Thoughts?

jannyHou:
@bajtos 
> to require callers of patch/replace operations to provide _rev in the data

A block of this idea is, user cannot get `rev` by `find()` or `findById()` since method `fromDB` deletes it:
https://github.com/strongloop/loopback-connector-cloudant/blob/master/lib/cloudant.js#L126-L127

My suggestion is adding a config option `includeRev` or simply `rev`:
  - if `true`, returns `rev` of data, don't delete it
  - if `false`, delete it as our current implementation. And make default as `false`

Then we can choose which signature of `insert()` to use according to whether `rev` provided. 
The reference for function `insert` in nodejs-cloudant: https://github.com/cloudant-labs/cloudant-nano#dbinsertdoc-params-callback

jannyHou:
Closing this pr since we agree on https://github.com/strongloop/loopback-connector-cloudant/issues/55

Issue #18, opened 2016-05-27
Cannot create data source "cloudant"


krsyoung:
Hey, I'm trying to connect my Loopback application to Cloudant but I am not having much luck.  I used the `slc loopback:datasource` command to add a Cloudant datasource, however, when I start my application I get:

```
chriss-mbp:project chrisyoung$ npm start

> project@1.0.0 start project
> node server/server.js

project/node_modules/loopback/lib/application.js:238
    throw err;
    ^

TypeError: Cannot create data source "cloudant": Cannot initialize connector "cloudant": Cannot set property 'use' of undefined
    at Cloudant (project/node_modules/loopback-connector-cloudant/node_modules/cloudant/cloudant.js:128:26)
    at new Cloudant (project/node_modules/loopback-connector-cloudant/lib/cloudant.js:52:19)
    at Object.exports.initialize (project/node_modules/loopback-connector-cloudant/lib/cloudant.js:21:18)
    at DataSource.setup (project/node_modules/loopback-datasource-juggler/lib/datasource.js:338:19)
    at new DataSource (project/node_modules/loopback-datasource-juggler/lib/datasource.js:114:8)
    at Registry.createDataSource (project/node_modules/loopback/lib/registry.js:349:12)
    at dataSourcesFromConfig (project/node_modules/loopback/lib/application.js:430:19)
    at EventEmitter.app.dataSource (project/node_modules/loopback/lib/application.js:228:14)
    at project/node_modules/loopback-boot/lib/executor.js:178:9
    at project/node_modules/loopback-boot/lib/executor.js:269:5
```

Here are the dependencies in my `package.json`:

``` json
  "dependencies": {
    "cfenv": "^1.0.3",
    "compression": "^1.6.2",
    "cors": "^2.7.1",
    "loopback": "^2.28.0",
    "loopback-boot": "^2.18.1",
    "loopback-component-explorer": "^2.5.0",
    "loopback-connector-cloudant": "^1.0.11",
    "loopback-datasource-juggler": "^2.46.0",
    "morgan": "^1.7.0",
    "newrelic": "^1.28.0",
    "serve-favicon": "^2.3.0",
    "winston": "^2.2.0",
    "winston-loggly": "^1.3.1"
  }
```

And my `datasources.json`:

``` json
{
  "db": {
    "name": "db",
    "connector": "memory"
  },
  "cloudant": {
    "url": "https://xyz.cloudant.com/apih-tk",
    "database": "project",
[...]

PR #60, opened 2016-11-23
Explain uncaught url error in doc

jannyHou:
Connect to strongloop/loopback-connector-cloudant#18

Add explanation to warn users an invalid url string will result in uncaught error and fail their app.

PTAL @strongloop/squad-apex . Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,7 @@ password  | String | Cloudant password
 url       | String | Cloudant URL containing both username and password
 modelIndex | String | Specify the model name to document mapping, defaults to 'loopback\_\_model\_\_name'
 
+**NOTE: Please be aware that the url will override username/password.**
 
 ### Model Specific Configuration
 


README.md
```diff
@@ -68,6 +68,12 @@ password  | String | Cloudant password
 url       | String | Cloudant URL containing both username and password
 modelIndex | String | Specify the model name to document mapping, defaults to 'loopback\_\_model\_\_name'
 
+**NOTE: Please be aware of that the URL will override username/password and must follow a [right format](https://github.com/cloudant/nodejs-cloudant#initialization)**:
```
kjdelisle: I would omit mentioning the format of the URL; the required format is a result of the URI standard.

Additionally, I would *not* mention the downstream issue in our comments, since it will ideally be fixed soon. Putting those comments in will just require removal later.

kjdelisle: Maybe just mention that the URL will override user/pass?

README.md
```diff
@@ -68,6 +68,7 @@ password  | String | Cloudant password
 url       | String | Cloudant URL containing both username and password
 modelIndex | String | Specify the model name to document mapping, defaults to 'loopback\_\_model\_\_name'
 
+**NOTE: Please be aware of that the url will override username/password.**
```
kjdelisle: Remove the "of'":P

Issue #56, opened 2016-11-16
Rate limits from new IBM Bluemix Cloudant plans are not managed by the connector


l2fprod:
Cloudant in Bluemix has plans with rate limits. It used to have a "Shared" plan with a limit of storage but the new plans have introduced read and write per second limits:
https://www.ibm.com/blogs/bluemix/2016/09/new-cloudant-lite-standard-plans-are-live-in-bluemix-public/

The Cloudant nodejs package has been updated to cope with the new limits and automatically retry. This is implemented with a Cloudant "retry" plugin:
https://github.com/cloudant/nodejs-cloudant#request-plugins

The loopback-connnector-cloudant should be updated to use the latest Cloudant nodejs package and support the configuration of the retry plugin.

Typically I'm getting http error 429 (the Cloudant error code when you reach the rate limit) and messages like:

```
Error: You've exceeded your current limit of 5 requests per second for query class. Please try later.
    at Request._callback (/Users/fred/dev/api/node_modules/nano/lib/nano.js:248:15)
    at Request.self.callback (/Users/fred/dev/api/node_modules/request/request.js:187:22)
    at emitTwo (events.js:106:13)
    at Request.emit (events.js:191:7)
    at Request.<anonymous> (/Users/fred/dev/api/node_modules/request/request.js:1044:10)
    at emitOne (events.js:96:13)
    at Request.emit (events.js:188:7)
    at IncomingMessage.<anonymous> (/Users/fred/dev/api/node_modules/request/request.js:965:12)
    at emitNone (events.js:91:20)
```

l2fprod:
I tried a local fix https://github.com/l2fprod/loopback-connector-cloudant/commit/f342236322aa0f959d9a423a8238387c8ba4c8c7 enabling the retry plugin and obviously it solves the issue. Ideally this should be built-in into the connector with configuration from the datasource.

xpqz:
It would be great if StrongLoop's Cloudant connector could wrap Cloudant's nodejs library. The Shared plan has been deprecated in favour of the new rate-capped Lite & Standard plans, and at the moment SL dies when used against the Lite plan. The Lite plan is intended to serve as a free-forever taster/development tier but applications using that need to be mindful of the caps. Cloudant's library has configurable backoff&retry handling through its `retry` plugin. 

Stefan Kruger, IBM Cloudant.

superkhau:
[...]

PR #69, opened 2017-01-05
cloudant: allow passthrough of settings

kjdelisle:
### Description
Arbitrary properties of the settings object will now be passed
to the driver to allow configuration of nodejs-cloudant settings.

#### Related issues
#56

### Checklist

### DIFF
--- a/lib/cloudant.js
+++ b/lib/cloudant.js
@@ -41,6 +41,8 @@ exports.initialize = function(ds, cb) {
  * @constructor
  */
 function Cloudant(settings, ds) {
+  // Injection for tests
+  var CloudantDriver = settings.Driver || Driver;
   debug('Cloudant constructor settings: %j', settings);
   Connector.call(this, 'cloudant', settings);
   this.debug = settings.debug || debug.enabled;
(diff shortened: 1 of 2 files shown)


kjdelisle:
Still need to create some tests, just waiting to see what the results are on CI.

test/cloudant.test.js
```diff
@@ -111,3 +113,25 @@ describe('cloudant connector', function() {
...
+      ds.settings.Driver = function(options) {
+        result = options;
+      };
+      ds.settings.foobar = {
+        foo: 'bar',
+      };
+      ds.settings.plugin = 'whack-a-mole';
+      var connector = Cloudant.initialize(ds, function(err) {
+        should.not.exist(err);
+        should.exist(result.foobar);
```
b-admike: where is result coming from? the callback function?

kjdelisle: result is being set in the injected fake for Driver:
```
var result = {};
ds.settings = ds.settings || {};
ds.settings.Driver = function(options) {
  result = options;
};
```

lib/cloudant.js
```diff
@@ -49,11 +51,17 @@ function Cloudant(settings, ds) {
...
     ' AND "password" required');
   }
-  this.creds = settings.url || {
-    account: settings.username,
-    password: settings.password,
-  };
-  this.cloudant = Driver(this.creds);
+
+  if (settings.url) {
+    this.options = settings.url;
```
jannyHou: I am checking if url will override settings.

jannyHou: @kjdelisle From cloudant document I think they still expect arbitrary properties like `plugin` to be passed in separately, not inside `url`:
e.g: the example in https://github.com/cloudant/nodejs-cloudant#the-promises-plugins
`var cloudant = Cloudant({url: myurl, plugin:'promises'});`

so with `url` we still need `_.merge(this.options, settings);`.

kjdelisle: The current code doesn't pass an object if url is given:
```
this.creds = settings.url || { //etc.. };
```

kjdelisle: If always constructing an object and just appending the properties is valid, I can just do that.

PR #72, opened 2017-01-13
Fix updateAll for model instance as input

### LINKED ISSUE #71: udpateAll( ) fails if input data object is a LoopBack model instance
### Description of feature (or steps to reproduce if bug)
When `updateAll()` is invoked with loopback `model` instance as payload, it fails with complaining: `Bad special document member: __cachedRelations` and similar. 
Other NoSQL connector i.e. mongo connector, AFAIK, supports `model` instance as payload.

gunjpan:
invoking updateAll() with input data in loopback
model instance format was causing failures as
it was not converted to canonical presentation.

This fix resolves it by checking if input data
is a model instance and if it is, convert it to
a plain JSON object for update.

Fixes #71

#### Related issues

- None

### Checklist

### DIFF
--- a/lib/cloudant.js
+++ b/lib/cloudant.js
@@ -514,6 +514,8 @@ Cloudant.prototype.updateAttributes = function(model, id, data, options, cb) {
   var mo = self.selectModel(model);
   mo.db.get(id, function(err, doc) {
     if (err) return cb(err, doc);
+    
+    data = getPlainJSONData.call(self, model, data);
     _.merge(doc, data);
     self.create(model, doc, options, function(err) {
       cb(err, self.fromDB(model, mo, doc));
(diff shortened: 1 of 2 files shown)


gunjpan:
Linter failing :man_facepalming:

lib/cloudant.js
```diff
@@ -522,6 +524,19 @@ Cloudant.prototype.updateAttributes = function(model, id, data, options, cb) {
...
+ * REST endpoint as remoting converts it to model if endpoint expects a
+ * model instance
+ */
+
+function getPlainJSONData(model, data) {
+  if (this._models[model] && data instanceof this._models[model].model) {
+    return data.toJSON();
+  }
+  return data;
+}
```
kjdelisle: It might be that I don't understand exactly what's happening here, but it seems to me like this function is the sort of thing that would happen in the base connector repository, and not in this particular one. How is this handled in other connectors?

gunjpan: AFAIK, this function i.e. parsing data from model instance is not present in the base connector and in other connectors, there is some sort of cleanup built-in for input data.
Thus, defined parsing of input data here in the connector.

gunjpan: @kjdelisle : :point_up: FYI

gunjpan:
@kjdelisle : Is this good to land?

kjdelisle:
LGTM

Issue #75, opened 2017-02-01
Append fieldType when sort with nested property 


jannyHou:
Related to https://github.com/strongloop/loopback/issues/517#issuecomment-275532100

cloudant sort needs specified fieldType, either :string or :number, details please check https://docs.cloudant.com/cloudant_query.html#sort-syntax

Now we already append the fieldType for first level properties but not nested ones. Therefore when sorting with nested property(e.g. [test in juggler](https://github.com/strongloop/loopback-datasource-juggler/pull/1239/files#diff-8fae24fa21c82e23f7f43ccdae4fcf90R636)), cloudant complains 
> Unspecified or ambiguous sort type. Try appending :number or :string to the sort field. address.city

jannyHou:
Solution:
- From the perspective of loopback, users should be able to query nested property without specifying its type, e.g. `{order: 'address.city DESC'}`, https://github.com/strongloop/loopback-connector-cloudant/pull/76 provides the default field type as `string`, and users can override it by appending their type in query, e.g. `{order: 'address.city:number DESC'}`

- From the perspective of cloudant, users are responsible to specify the type if by default it's 'Unspecified or ambiguous sort type', in this case there is no fix on our end but we need to well document it.

jannyHou:
@strongloop/sq-lb-apex PTAL. Thanks.

jannyHou:
Closing since pr merged

jannyHou:
Fixed in PR#90

PR #76, opened 2017-02-01
Append field type for nested properties

jannyHou:
Connect to strongloop/loopback-connector-cloudant#75

Append field type for nested properties. Before the pr it only applies to first level property.
Test case is in juggler https://github.com/strongloop/loopback-datasource-juggler/pull/1239/files#diff-8fae24fa21c82e23f7f43ccdae4fcf90R635


0candy:
@jannyHou Can you add a unit test?

jannyHou:
@0candy The test is in another juggler pr: https://github.com/strongloop/loopback-datasource-juggler/pull/1239/files#diff-8fae24fa21c82e23f7f43ccdae4fcf90R635

Issue #78, opened 2017-02-01
use $elemMatch when query on elements in an array field


jannyHou:
$elemMatch: Matches and returns all documents that contain an array field with at least one element that matches all the specified query criteria.

Use case:
*seed data*
```js
{
      seq: 0,
      name: 'John Lennon',
      friends: [
        {name: 'Paul McCartney'},
        {name: 'George Harrison'},
        {name: 'Ringo Starr'},
      ],
    },
```
*wrong query*
`{where: {friends.name: 'Ringo Starr'}}`
*right query*
`{where: {friends: {$elemMatch: {name: 'Ringo Starr'}}}}`

jannyHou:
Reopen this issue since the previous fix only works when model is created in code.

Several problem with the previous pr, which need to be fixed:

- [ ] The filter only works with apis in code but doesn't work with explorer
- [ ] The previous pr is based on an assumption that the connector can get all dataType of nested properties. If the model is created in code, like what test case does, all dataType are available, while when creating model by cli, loading it from the model.json file, for some reason deep nested property's type are undefined.
- [ ] The previous pr doesn't catch some undefined situation.

jannyHou:
UPDATE: 
Before solving the bugs mentioned in https://github.com/strongloop/loopback-connector-cloudant/issues/78#issuecomment-284454583 we are not sure should we wrap the array property by detecting property type in loopback, or add a detailed doc to tell the user to include `elemMatch` in their filter?

cc @raymondfeng

jannyHou:
fixed in PR#90

PR #80, opened 2017-02-03
Build selector with array type data

jannyHou:
connect to strongloop/loopback-connector-cloudant#78

When filter against array data, e.g. `friends.name` in the example below is `Array`, the current query is:
 `{where: {friends.name: 'Ringo Starr'}}`
while the *right query* is: 
`{where: {friends: {$elemMatch: {name: 'Ringo Starr'}}}}`

This PR wraps array filter with `$elemMatch`.

---

Test case is in juggler: https://github.com/strongloop/loopback-datasource-juggler/pull/1239/files#diff-8fae24fa21c82e23f7f43ccdae4fcf90R646

### DIFF
--- a/lib/cloudant.js
+++ b/lib/cloudant.js
@@ -365,6 +365,39 @@ Cloudant.prototype.buildSelector = function(model, mo, where) {
     } else {
       query[k] = cond;
     }
+    cond = query[k];
+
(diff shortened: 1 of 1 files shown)


jannyHou:
@strongloop/sq-lb-apex PTAL. Thanks.

Issue #73, opened 2017-01-16
TypeError: Cannot read property 'length' of undefined loopback-connector-cloudant/lib/cloudant.js:230


vinodk21:
Hi ,

Application is crashing frequently because of the following error. (loopback-connector-cloudant 1.0.11 ). 

ERR /home/vcap/app/node_modules/loopback-connector-cloudant/lib/cloudant.js:230
2017-01-16T15:38:39.42+0530 [App/0]      ERR       for (var i = 0; i < docs.length; i++) {
2017-01-16T15:38:39.42+0530 [App/0]      ERR                               ^
2017-01-16T15:38:39.42+0530 [App/0]      ERR TypeError: Cannot read property 'length' of undefined
2017-01-16T15:38:39.42+0530 [App/0]      ERR     at include (/home/vcap/app/node_modules/loopback-connector-cloudant/lib/cloudant.js:230:31)
2017-01-16T15:38:39.42+0530 [App/0]      ERR     at Request._callback (/home/vcap/app/node_modules/loopback-connector-cloudant/node_modules/cloudant/node_modules/cloudant-nano/lib/nano.js:216:16)
2017-01-16T15:38:39.42+0530 [App/0]      ERR     at Request.self.callback (/home/vcap/app/node_modules/loopback-connector-cloudant/node_modules/cloudant/node_modules/request/request.js:186:22)
2017-01-16T15:38:39.42+0530 [App/0]      ERR     at emitTwo (events.js:87:13)
2017-01-16T15:38:39.42+0530 [App/0]      ERR     at Request.emit (events.js:172:7)
2017-01-16T15:38:39.42+0530 [App/0]      ERR     at emitOne (events.js:77:13)
2017-01-16T15:38:39.42+0530 [App/0]      ERR     at Request.emit (events.js:169:7)
2017-01-16T15:38:39.42+0530 [App/0]      ERR     at IncomingMessage.<anonymous> (/home/vcap/app/node_modules/loopback-connector-cloudant/node_modules/cloudant/node_modules/request/request.js:980:12)
2017-01-16T15:38:39.42+0530 [App/0]      ERR     at IncomingMessage.g (events.js:260:16)
2017-01-16T15:38:39.42+0530 [App/0]      ERR     at emitNone (events.js:72:20)
2017-01-16T15:38:39.42+0530 [App/0]      ERR     at endReadableNT (_stream_readable.js:923:12)
[...]

PR #85, opened 2017-03-02
Catch undefined docs

jannyHou:
#### Related issues
https://github.com/strongloop/loopback-connector-cloudant/issues/73

### Checklist

- [x] Catch undefined docs as a workaround
- [x] Add debug info to help us debug

### DIFF
--- a/lib/cloudant.js
+++ b/lib/cloudant.js
@@ -270,6 +270,15 @@ function findRecursive(mo, query, docs, include, cb) {
   mo.db.find(query, function(err, rst) {
     debug('Cloudant.prototype.all (findRecursive) results: %j %j', err, rst);
     if (err) return cb(err);
+    // work around for issue
+    // https://github.com/strongloop/loopback-connector-cloudant/issues/73
+    if (!rst.docs) {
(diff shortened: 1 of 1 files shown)


lib/cloudant.js
```diff
@@ -270,6 +270,12 @@ function findRecursive(mo, query, docs, include, cb) {
   mo.db.find(query, function(err, rst) {
     debug('Cloudant.prototype.all (findRecursive) results: %j %j', err, rst);
     if (err) return cb(err);
+    // work around for issue 
+    // https://github.com/strongloop/loopback-connector-cloudant/issues/73
+    if (!rst.docs) {
+      debug('Cloudant.prototype.all (findRecursive) query: %j', query);
+      return cb('The docs in results cannot be undefined');
```
kjdelisle: This should be an Error object, not a string!
Something like:
```
// Import util up above with var util = require('util');
var queryView = util.inspect(query, 4);
debug('findRecursive query: %s', queryView);
var errMsg = util.format('No documents returned for query: %s', queryView);
return cb(new Error(msg));
```

PR #90, opened 2017-03-14
Support nested property filter

jannyHou:
#### Related issues

Parent issue: strongloop/loopback#517
Sub-issue in cloudant:
- support array type: https://github.com/strongloop/loopback-connector-cloudant/issues/78
- append sorting data type: https://github.com/strongloop/loopback-connector-cloudant/issues/75
pr for issue#78 is reverted because the algorithm need redesign.

This pr includes implementation of issue#78 and fix for issue#75 
Related test cases in juggler: https://github.com/strongloop/loopback-datasource-juggler/blob/master/test/basic-querying.test.js#L602

#### Background
Support array type: 
  - given a nested field like `field1.field2.field3.field4`, cloudant searches datatype for each field, and insert a `$elemMatch` after if it's an array. e.g. if field2 is an array, the new field will be `field1.field2.$elemMatch.field3.field4`.
  - cloudant db only takes in json format selector when $elemMatch included, so connector unfolds string field to a json one if it's a nested property.
  - the algorithm ensures that in the middle if any datatype is not defined in modelDef or the searching type is `undefined`, user can still provide a complete cloudant selector in query and the request will send the original query. 

Append sorting data type:
  - we assume the nested sorting field doesn't contain array type.

#### Test scenarios(used when verify story)

Check https://gist.github.com/jannyHou/cae4d0b88525b47289a83c11e96b453b

CI is now failing due to connection failed. Related test cases pass on local.

### DIFF
--- a/test/cloudant.test.js
+++ b/test/cloudant.test.js
@@ -8,11 +8,11 @@
 
 'use strict';
 
+require('./init.js');
 var Cloudant = require('../lib/cloudant');
 var _ = require('lodash');
 var should = require('should');
-var describe = require('./describe');
-var db, Product;
+var db, Product, CustomerSimple;
 
 describe('cloudant connector', function() {
   before(function(done) {
@@ -24,12 +24,42 @@ describe('cloudant connector', function() {
       price: {type: Number},
     }, {forceId: false});
 
+    // CustomerSimple means some nested property defs are missing in modelDef,
+    // tests for CustomerSimple are created to make sure the typeSearch algorithm
+    // won't crash when iterating
+    CustomerSimple = db.define('CustomerSimple', {
+      name: {
+        type: String,
+      },
+      seq: {
+        type: Number,
+      },
+      address: {
+        street: String,
+        state: String,
+        zipCode: String,
+        tags: [],
+      },
+      friends: [],
+      favorate: {
+        labels: [
+          {label: String},
+        ],
+      },
+    });
     Product.destroyAll(function(err) {
-      done();
+      CustomerSimple.destroyAll(function(err) {
+        done();
+      });
     });
   });
 
   describe('replaceOrCreate', function() {
+    after(function cleanUpData(done) {
+      Product.destroyAll(function(err) {
+        done();
+      });
+    });
     it('should replace a model instance if the passing key already exists',
       function(done) {
         Product.create({
@@ -73,6 +103,11 @@ describe('cloudant connector', function() {
   });
 
   describe('replaceById', function() {
+    after(function cleanUpData(done) {
+      Product.destroyAll(function(err) {
+        done();
+      });
+    });
     it('should replace the model instance if the provided key already exists',
       function(done) {
         Product.create({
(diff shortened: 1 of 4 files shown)


test/cloudant.test.js
```diff
@@ -113,7 +143,7 @@ describe('cloudant connector', function() {
       });
   });
 
-  describe('updateAll and updateAttributes', function() {
+  describeskip('updateAll and updateAttributes', function() {
```
raymondfeng: Why skip?

jannyHou: @raymondfeng currently all test cases are skipped on CI, but will run when testing on local. So does this one, I name it as describeskip but it will still run in local test.

We have an issue opened to fix all test cases but haven't got a chance to do it, and we are planning to clean up tests together when refactor the code to support couchdb.

I try to enable test cases related in this pr on CI while still keep others skipped.

jannyHou: @raymondfeng I enable `updateAll and updateAttributes` and make it pass on local too.

test/cloudant.test.js
```diff
@@ -8,11 +8,12 @@
 
 'use strict';
 
+require('./init.js');
 var Cloudant = require('../lib/cloudant');
 var _ = require('lodash');
 var should = require('should');
-var describe = require('./describe');
-var db, Product;
+var describeskip = require('./describe');
```
raymondfeng: This is not used

jannyHou:
@raymondfeng comments applied, PTAL again. Thank you.
Unfortunately we still need some time to setup CI, the tests pass on local so I think the pr is ready to review :
```
jannyhous-mbp:loopback-connector-cloudant jannyhou$ mocha test/cloudant.test.js


  cloudant connector
    replaceOrCreate
      ✓ should replace a model instance if the passing key already exists (365ms)
    replaceById
      ✓ should replace the model instance if the provided key already exists (177ms)
    updateAll and updateAttributes
      ✓ accepts loopback model instance as input data for update (165ms)
    nested property
      missing in modelDef
        ✓ returns result when nested property is notan array type (141ms)
        ✓ returns null when first level property is array
        ✓ returns result when first level property is array type and $elemMatch provided (65ms)
        ✓ returns null when multi-level nested property contains array type
        ✓ returns result when multi-level nested property contains array type and $elemMatch provided (88ms)
        ✓ returns error missing data type when sorting
        ✓ returns result when sorting type provided (46ms)
      defined in modelDef
        ✓ returns result when complete query of multi-level nested property provided (52ms)

  cloudant constructor
    ✓ should allow passthrough of properties in the settings object
    ✓ should pass the url as an object property


  13 passing (2s)


```

jannyHou:
@slnode test please

jannyHou:
Some update about CI:
CI uses bluemix cloudant server, which has a limit of request per second. So we have to increase the  mocha timeout. 
Testing with docker cloudant image pass immediately(10000ms would be enough), while with bluemix one the delay is very obvious.

jannyHou:
@slnode test please

jannyHou:
@slnode test please

jannyHou:
@slnode test please

jannyHou:
@slnode test please

PR #95, opened 2017-04-06
add url config example

b-admike:
Update readme to have example for
datasource config using url field.
Also, specify limitations for our
cloudant driver.

#### Related issues

#18 

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,11 +37,26 @@ look something like this:
 "mydb": {
   "name": "mydb",
   "connector": "cloudant",
+  "url": "https://<username>:<password>@<host>"
+  "database": "test"
+}
+```
+or
+
+```
+"mydb": {
+  "name": "mydb",
+  "connector": "cloudant",
   "username": "XXXX-bluemix",
   "password": "YYYYYYYYYYYY",
   "database": "test"
 }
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -37,12 +37,27 @@ look something like this:
...
+"mydb": {
+  "name": "mydb",
+  "connector": "cloudant",
   "username": "XXXX-bluemix",
   "password": "YYYYYYYYYYYY",
   "database": "test"
 }
 ```
 
+only specify the `username` and `password` fields if you are using the root/admin user for the cloudant server which has the same
```
0candy: `Only`

README.md
```diff
@@ -37,12 +37,27 @@ look something like this:
...
+string as the hostname for the cloudant server, because the cloudant driver used by the connector appends `.cloudant.com` to the
+`username` field when `username` and `password` fields are specified. Therefore, it is good practice to use the `url` field instead
```
jannyHou: > Therefore, it is good practice to use the `url` field instead

Probably change to 
> Therefore, it is good practice to use the `url` field if your host is not `username.cloudant.com`

PR #97, opened 2017-04-10
Increase test timeout

ssh24:


ssh24:
@slnode test please

jannyHou:
@ssh24 I will create another PR with timestamp, see where does that super long time consuming happens.

jannyHou:
I think we are good with the current timeout after changing to the stable server.

jannyHou:
@ssh24 Since we get a new stable server for testing, shall we close this PR? Thanks.

ssh24:
Yes @jannyHou Closing this since we switched to the new server.

PR #98, opened 2017-04-10
call driver inside connect

b-admike:
### Description
call our cloudant driver inside connect function when connector initializes and strip the path of the url and use it as the database name before passing it to our driver because of https://github.com/cloudant/nodejs-cloudant/issues/181.

#### Related issues

connected to https://github.com/strongloop-internal/scrum-apex/issues/38

### Checklist

### DIFF
--- a/lib/cloudant.js
+++ b/lib/cloudant.js
@@ -88,7 +88,15 @@ Cloudant.prototype.connect = function(cb) {
   self.CloudantDriver(self.options, function(err, nano) {
     if (err) return cb(err);
     self.cloudant = nano;
-    return cb(err, self.cloudant);
+    if (self.options.database) {
+      // check if database exists
+      self.cloudant.db.get(self.options.database, function(err) {
+        if (err) return cb(err);
+        return cb(err, self.cloudant);
+      });
+    } else {
+      return cb(err, self.cloudant);
+    }
   });
 };
 
--- a/lib/cloudant.js
+++ b/lib/cloudant.js
@@ -11,6 +11,7 @@ var Driver = require('cloudant');
 var assert = require('assert');
 var debug = require('debug')('loopback:connector:cloudant');
 var async = require('async');
+var url = require('url');
 var util = require('util');
 var _ = require('lodash');
 
@@ -42,7 +43,7 @@ exports.initialize = function(ds, cb) {
  */
 function Cloudant(settings, ds) {
   // Injection for tests
-  var CloudantDriver = settings.Driver || Driver;
+  this.CloudantDriver = settings.Driver || Driver;
   debug('Cloudant constructor settings: %j', settings);
   Connector.call(this, 'cloudant', settings);
   this.debug = settings.debug || debug.enabled;
@@ -51,14 +52,12 @@ function Cloudant(settings, ds) {
     throw new Error('Invalid settings: "url" OR "username"' +
     ' AND "password" required');
   }
-
   this.options = _.merge({}, settings);
   // If settings.url is not set, then setup account/password props.
   if (!this.options.url) {
     this.options.account = settings.username;
     this.options.password = settings.password;
   }
-  this.cloudant = CloudantDriver(this.options);
   this.pool = {};
 }
 
@@ -75,7 +74,22 @@ Cloudant.prototype.getTypes = function() {
  */
 Cloudant.prototype.connect = function(cb) {
   debug('Cloudant.prototype.connect');
-  cb(null, this.cloudant);
+  var self = this;
+ // strip db name if defined in path of url before
+  // sending it to our driver
+  if (self.options.url) {
+    var parsedUrl = url.parse(self.options.url);
+    if (parsedUrl.path && parsedUrl.path !== '/') {
+      self.options.url = self.options.url.replace(parsedUrl.path, '');
+      if (!self.options.database)
+        self.options.database = parsedUrl.path.split('/')[1];
+    }
+  }
+  self.CloudantDriver(self.options, function(err, nano) {
+    if (err) return cb(err);
+    self.cloudant = nano;
+    return cb(err, self.cloudant);
+  });
 };
 
 /**
--- a/test/cloudant.test.js
+++ b/test/cloudant.test.js
@@ -8,6 +8,7 @@
 require('./init.js');
 var Cloudant = require('../lib/cloudant');
 var _ = require('lodash');
+var url = require('url');
 var should = require('should');
 var db, Product, CustomerSimple;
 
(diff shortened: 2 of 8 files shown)


b-admike:
@slnode test please

lib/cloudant.js
```diff
@@ -75,7 +75,42 @@ Cloudant.prototype.getTypes = function() {
  */
 Cloudant.prototype.connect = function(cb) {
   debug('Cloudant.prototype.connect');
-  cb(null, this.cloudant);
+  var self = this;
+  // check if cloudant server is up
+  if (self.options.url) {
+    request.get(self.options.url, function(err, res, body) {
```
ssh24: Do we not care about the `response` or the `body` returned back at all?

lib/cloudant.js
```diff
@@ -75,7 +75,42 @@ Cloudant.prototype.getTypes = function() {
...
+    var opts = {
+      auth: {
+        user: self.options.username,
+        pass: self.options.password,
+        sendImmediately: true,
+      },
+      uri: self.options.username + '.cloudant.com',
+      port: self.options.port || '',
+    };
+    request.get(opts, function(err, res, body) {
```
ssh24: Same here. Do we not care about the `response` or the `body`?

test/cloudant.test.js
```diff
@@ -324,6 +324,26 @@ describe('cloudant constructor', function() {
...
+      result = options;
+    };
+    ds.settings.url = 'https://fakeuser:fakepass@definitelynotreal.cloudant.com:8080/some/random/path';
+    var connector = Cloudant.initialize(ds, function(err) {
+      should.exist(err);
+      should.exist(result.url);
+      result.url.should.equal('https://fakeuser:fakepass@definitelynotreal.cloudant.com:8080');
+      result.database.should.not.equal('some');
+    });
+    ds.settings.database = undefined;
```
jannyHou: If `Cloudant.initialize()` is a async function, I think line#341-line#345 should be moved into line#335's callback function?

lib/cloudant.js
b-admike: `response` yes it is used to send back the err i.e. `res.errMsg`, but we don't care about `body`

lib/cloudant.js
b-admike: `response` yes it is used to send back the err i.e. `res.errMsg`, but we don't care about `body`

kjdelisle: Can we ensure that both the error and the appropriate HTTP status code are returned as part of the error message?

lib/cloudant.js
ssh24: Could we remove the `body` parameter then?

kjdelisle: Can we ensure that both the error and the appropriate HTTP status code are returned as part of the error message?

test/cloudant.test.js
```diff
@@ -324,6 +324,26 @@ describe('cloudant constructor', function() {
...
+    var result = {};
+    ds.settings.Driver = function(options) {
+      result = options;
+    };
+    ds.settings.url = 'https://fakeuser:fakepass@definitelynotreal.cloudant.com:8080/some/random/path';
+    var connector = Cloudant.initialize(ds, function(err) {
+      should.exist(err);
+      should.exist(result.url);
+      result.url.should.equal('https://fakeuser:fakepass@definitelynotreal.cloudant.com:8080');
+      result.database.should.not.equal('some');
```
kjdelisle: There should be several tests here for various errors and status codes.
- 0 for no response
- 400 for bad request
- 404 for not found
- 401 for Unauthorized
- 403 for Forbidden
etc...

lib/cloudant.js
```diff
@@ -75,7 +75,42 @@ Cloudant.prototype.getTypes = function() {
...
+      if (err) cb(res.errMsg);
+    });
+  }
+  // strip db name if defined in path of url before
+  // sending it to our driver
+  if (self.options.url) {
```
jannyHou: Looks like this block(L#100-L#113) and the previous one(L#80-L#97) will execute in parallel. 
Maybe wrap this block into a function, e.g. 'function getDriverInstance()', and call it after L#82 and L#95?

lib/cloudant.js
```diff
@@ -75,7 +75,42 @@ Cloudant.prototype.getTypes = function() {
  */
 Cloudant.prototype.connect = function(cb) {
   debug('Cloudant.prototype.connect');
-  cb(null, this.cloudant);
+  var self = this;
+  // check if cloudant server is up
+  if (self.options.url) {
+    request.get(self.options.url, function(err, res, body) {
+      if (err) cb(res.errMsg);
```
jannyHou: `if (err) return cb(res.errMsg);`

lib/cloudant.js
```diff
@@ -75,7 +75,42 @@ Cloudant.prototype.getTypes = function() {
...
+      auth: {
+        user: self.options.username,
+        pass: self.options.password,
+        sendImmediately: true,
+      },
+      uri: self.options.username + '.cloudant.com',
+      port: self.options.port || '',
+    };
+    request.get(opts, function(err, res, body) {
+      if (err) cb(res.errMsg);
```
jannyHou: same here: `if (err) return cb(res.errMsg);`

test/cloudant.test.js
b-admike: I removed the ping code in the previous commit because calling the driver in the callback of the `request.get` would give us the same issue as using the async version of our driver, so this is not applicable.

test/cloudant.test.js
b-admike: done

lib/cloudant.js
b-admike: I removed the ping code in the previous commit because calling the driver in the callback of the `request.get` would give us the same issue as using the async version of our driver (which is not getting the cloudant connection object back in time when we use it further down the line).

lib/cloudant.js
b-admike: Doing that made me realize that we won't get in the callback for `request.get` calls in time for using the cloudant connection object down the line, so I have removed it altogether.

lib/cloudant.js
b-admike: I removed the ping code altogether because calling the driver in the callback of the request.get would give us the same issue as using the async version of our driver

lib/cloudant.js
```diff
@@ -75,7 +74,23 @@ Cloudant.prototype.getTypes = function() {
  */
 Cloudant.prototype.connect = function(cb) {
   debug('Cloudant.prototype.connect');
-  cb(null, this.cloudant);
+  var self = this;
+ // strip db name if defined in path of url before
```
jannyHou: indention: one more space

b-admike:
@jannyHou the reason I removed the ping is for two reasons:
1) it does not return in time to call our cloudant driver and get the cloudant connection object before it is used again down the line
2) it happens later if you have an invalid config and fails accordingly

lib/cloudant.js
```diff
@@ -75,7 +74,23 @@ Cloudant.prototype.getTypes = function() {
...
+  if (self.options.url) {
+    var parsedUrl = url.parse(self.options.url);
+    if (parsedUrl.path && parsedUrl.path !== '/') {
+      self.options.url = self.options.url.replace(parsedUrl.path, '');
+      if (!self.options.database)
+        self.options.database = parsedUrl.path.split('/')[1];
+    }
+  }
+  try {
+    self.cloudant = self.CloudantDriver(self.options);
```
jannyHou: `self.CloudantDriver()` is a async function, could you try 
```js
self.cloudant = self.CloudantDriver(self.options, function (err, nano, pong) {
  if (err) return cb (err);
  cb (null, nano);
});

```
?

jannyHou: @b-admike sorry, try this:
```js
var connection = self.CloudantDriver(self.options, function (err, nano, pong) {
  if (err) return cb (err);
  self.cloudant = connection;
  cb (null, connection);
});
```

lib/cloudant.js
```diff
@@ -75,7 +74,23 @@ Cloudant.prototype.getTypes = function() {
  */
 Cloudant.prototype.connect = function(cb) {
   debug('Cloudant.prototype.connect');
-  cb(null, this.cloudant);
+  var self = this;
```
jannyHou:
```js
if (this.cloudant) {
    if (callback) {
      process.nextTick(function() {
        callback(null, self.cloudant);
      });
    }
  } else {
  // url strip
  var connection = self.CloudantDriver(self.options, function (err, nano, pong) {
  if (err) return cb (err);
  self.cloudant = connection_or_nano;
  cb (null, connection_or_nano);
});

}
```

b-admike:
@slnode test please

lib/cloudant.js
```diff
@@ -75,7 +74,30 @@ Cloudant.prototype.getTypes = function() {
...
+      if (!self.options.database)
+        self.options.database = parsedUrl.path.split('/')[1];
+    }
+  }
+  self.CloudantDriver(self.options, function(err, nano) {
+    if (err) return cb(err);
+    self.cloudant = nano;
+    if (self.options.database) {
+      // check if database exists
+      self.cloudant.db.get(self.options.database, function(err) {
```
jannyHou: @b-admike do we have to check `database` exist? I think we can leave the server to return error if user calls an api against a non-existing database.

b-admike:
@jannyHou I added it because I wanted to make sure when users start their apps, they have their config correct before making any requests. Our drivers ping function passes an empty string for dbName all the time (See L#271 in `cloudant.js`), so maybe we can make the change there. However, I understand most connectors don't do this so I don't mind removing it if the pros are more than the cons.

b-admike:
@slnode test please

kjdelisle:
Please be aware that this function also represents a breaking change.

PR #93, opened 2017-04-04
Add doc for filter and order

jannyHou:
Connect to https://github.com/strongloop/loopback-connector-cloudant/pull/90

Add more docs to explain filter and order in cloudant.

Issue #92, opened 2017-03-28
automigrate bug -- document conflict error


jannyHou:
**BUG**

- with empty db, or db doesn't have the model's document
- use `db.define()` to create model
- run `db.automigrate()` 
- it complains conflict error, since the automigrate function runs twice, once when connection established, once when the function get called. The conflict might raised by race condition(two model create request runs together)

jannyHou:
Turns out the problem is cloudant has its own `define` function that overrides the one in loopback-connector. It doesn't only add a model to datasource's `_model` object, but also **update** the index, which is exactly what `automigrate` does.

I remove `define` function in PR to solve the problem because in loopback-datasource-juggler and other loopback components like loopback-boot, it's used as a sync function, while update index in cloudant is an async function.

PR #91, opened 2017-03-27
Recover maxrows.test.js

jannyHou:
connect to strongloop/loopback-connector-cloudant#33

write request still runs into limit problem.

PR #102, opened 2017-04-18
Use define function in loopback-connector

jannyHou:
#### Related issues

strongloop/loopback-connector-cloudant#92

cloudant has its own `define` function that overrides the one in loopback-connector. It doesn't only add a model to datasource's `_model` object, but also **update** the index, which is exactly what `automigrate` does.

I remove `define` function in this PR because in loopback-datasource-juggler and other loopback components like loopback-boot, `define` is used as a sync function, while update index in cloudant is an async function. https://github.com/strongloop/loopback-connector-cloudant/issues/92 and https://github.com/strongloop/loopback-connector-cloudant/issues/93 are caused by this.

This will be a breaking change since user will have to call `db.automigrate` to migrate their models to database, cloudant won't automatically update the index in model's "define" phase.

#### Checklist

### DIFF
--- a/lib/cloudant.js
+++ b/lib/cloudant.js
@@ -874,19 +874,6 @@ Cloudant.prototype.selectModel = function(model, migrate) {
 };
 
 /**
- * Hook to be called by DataSource for defining a model
- *
- * @param {Object} modelDefinition The model definition
- */
-Cloudant.prototype.define = function(modelDefinition) {
-  var modelName = modelDefinition.model.modelName;
-  modelDefinition.settings = modelDefinition.settings || {};
-  this._models[modelName] = modelDefinition;
-  var mo = this.selectModel(modelName, true);
-  this.updateIndex(mo, modelName);
-};
-
-/**
  * Perform autoupdate for the given models. It basically calls db.index()
  *
  * @param {String[]} [models] A model name or an array of model names. If not
@@ -900,8 +887,15 @@ Cloudant.prototype.autoupdate = function(models, cb) {
 
 /**
  * Perform automigrate for the given models.
+ * Notes: Cloudant stores a Model as a design doc in database,
+ * `ddoc` is the doc's _id, auto-generated as 'lb-index-ddoc' + modelName,
+ * `name` is the index name, auto-generated as 'lb-index' + modelName.
+ * It does NOT store properties in the doc, which makes autoupdate actually
+ * does the same thing as automigrate.
+ * It does NOT destroy previous model instances if model exists, user needs
+ * to manually destroy them if necessary.
  *
- * @param {String[]} [models] A model name or an array of model names. If not
+ * @param {String|String[]} [models] A model name or an array of model names. If not
  * present, apply to all models
  * @param {Function} [cb] The callback function
  */
@@ -948,7 +942,9 @@ Cloudant.prototype.updateIndex = function(mo, modelName, cb) {
     },
   };
   /* eslint-enable camelcase */
-  debug('Cloudant.prototype.updateIndex modelName %j', modelName);
+  var indexView = util.inspect(idx, 4);
+  debug('Cloudant.prototype.updateIndex -- modelName %s, idx %s', modelName,
+    indexView);
   if (mo.modelSelector === null) {
     idx.index.selector[mo.modelView] = modelName;
   }
--- a/test/automigrate.test.js
+++ b/test/automigrate.test.js
@@ -1,0 +1,31 @@
+// Copyright IBM Corp. 2015,2016. All Rights Reserved.
+// Node module: loopback-connector-cloudant
+// This file is licensed under the Artistic License 2.0.
+// License text available at https://opensource.org/licenses/Artistic-2.0
+
+'use strict';
+var db, Foo, Bar;
+
+describe('cloudant imported features', function() {
+  before(function() {
+    require('./init.js');
+  });
+  it('automigrates models attached to db', function(done) {
+    db = getSchema();
+    Foo = db.define('Foo', {
+      name: {type: String},
+    });
+    Bar = db.define('Bar', {
+      id: {type: Number, index: true},
+    });
+    db.automigrate(function verifyMigratedModel(err) {
+      if (err) return done(err);
+      Foo.create({name: 'foo'}, function(err, r) {
+        if (err) return done(err);
+        r.should.not.be.empty();
+        r.name.should.equal('foo');
+        done();
+      });
+    });
+  });
+});
--- a/test/cloudant.test.js
+++ b/test/cloudant.test.js
@@ -3,9 +3,6 @@
 // This file is licensed under the Artistic License 2.0.
 // License text available at https://opensource.org/licenses/Artistic-2.0
 
-// Comment test cases to get CI pass,
-// will recover them when CI config done
-
 'use strict';
 
 require('./init.js');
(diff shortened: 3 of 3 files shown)


test/automigrate.test.js
```diff
@@ -0,0 +1,29 @@
...
+  });
+  it('automigrates models attached to db', function(done) {
+    db = getSchema();
+    Foo = db.define('Foo', {
+      name: {type: String},
+    });
+    Bar = db.define('Bar', {
+      id: {type: Number, index: true},
+    });
+    db.automigrate(function createSampleData() {
```
ssh24: do we need to assign the function a name? (i.e `createSampleData`)

test/cloudant.test.js
```diff
@@ -47,9 +47,13 @@ describe('cloudant connector', function() {
         ],
       },
     });
-    Product.destroyAll(function(err) {
-      CustomerSimple.destroyAll(function(err) {
-        done();
+
+    db.automigrate(function(err) {
```
ssh24: Regarding my last comment, we should add a function name here to make it consistent.

lib/cloudant.js
```diff
@@ -900,6 +887,13 @@ Cloudant.prototype.autoupdate = function(models, cb) {
 
 /**
  * Perform automigrate for the given models.
+ * Notes: Cloudant stores a Model as a design doc in database,
+ * `ddoc` is the doc's _id, auto-generated as 'lb-index-ddoc' + modelName,
+ * `name` is the index name, auto-generated as 'lb-index' + modelName.
+ * It does NOT store properties in the doc, which makes autoudpate actually
```
0candy: autoupdate typo

test/cloudant.test.js
```diff
@@ -47,9 +47,13 @@ describe('cloudant connector', function() {
...
+      if (err) return done(err);
+      Product.destroyAll(function(err) {
```
ssh24: Same here.

test/automigrate.test.js
jannyHou: ah good catch, will remove it.

test/cloudant.test.js
```diff
@@ -47,9 +47,13 @@ describe('cloudant connector', function() {
...
+        CustomerSimple.destroyAll(function(err) {
```
ssh24: Same here.

lib/cloudant.js
```diff
@@ -948,7 +942,7 @@ Cloudant.prototype.updateIndex = function(mo, modelName, cb) {
     },
   };
   /* eslint-enable camelcase */
-  debug('Cloudant.prototype.updateIndex modelName %j', modelName);
+  debug('Cloudant.prototype.updateIndex modelName %j idx', modelName, idx);
```
0candy: You need a second formatter to display the `idx` value.

test/cloudant.test.js
jannyHou: sounds good, I prefer to update all of them in https://github.com/strongloop/loopback-connector-cloudant/pull/91, it is a test case clean up pr.

ssh24: Thank you. Other than that LGTM

jannyHou:
@0candy good catch thanks, fixed.

lib/cloudant.js
```diff
@@ -948,7 +942,7 @@ Cloudant.prototype.updateIndex = function(mo, modelName, cb) {
     },
   };
   /* eslint-enable camelcase */
-  debug('Cloudant.prototype.updateIndex modelName %j', modelName);
+  debug('Cloudant.prototype.updateIndex modelName %j idx %j', modelName, idx);
```
kjdelisle: `debug('Cloudant.prototype.updateIndex -- modelName: %s, idx: %s', modelName, idx)`

If the modelName or idx are objects and not strings for some weird reason, you can use `require('util').inspect` to stringify them nicely.

test/automigrate.test.js
kjdelisle: I like the idea. It accurately describes what the function's purpose is.
(Of course, you should never take for granted that what the function _says_ it does is what it's actually doing, but it's still helpful.)

I would avoid naming the function definition on the right-hand side of the assignment operator if the left-hand was named (`Foo.prototype.createWidget = function createWidget() { ... }`), but this isn't one of those cases.

test/cloudant.test.js
kjdelisle: I would prefer that we make this change here, since _this only exists in a state that needs fixing_ because of this PR!
Let's continue naming these anonymous functions under test; it's a good pattern IMO.

test/cloudant.test.js
kjdelisle: :point_up:

test/cloudant.test.js
kjdelisle: ☝️

lib/cloudant.js
```diff
@@ -900,6 +887,13 @@ Cloudant.prototype.autoupdate = function(models, cb) {
...
  * Perform automigrate for the given models.
+ * Notes: Cloudant stores a Model as a design doc in database,
+ * `ddoc` is the doc's _id, auto-generated as 'lb-index-ddoc' + modelName,
+ * `name` is the index name, auto-generated as 'lb-index' + modelName.
+ * It does NOT store properties in the doc, which makes autoupdate actually
+ * does the same thing as automigrate.
+ * It does NOT destroy previous model instances if model exists, user needs
+ * to manually destroy them if necessary.
  *
  * @param {String[]} [models] A model name or an array of model names. If not
```
kjdelisle: While you're here, since you've updated the JSDoc description, can you also change the parameter type to be `{String|String[]}`? The description to the right of the parameter definition doesn't match the given type definition.

test/cloudant.test.js
jannyHou: cool, will add function names here.

jannyHou:
thanks @kjdelisle Changed applied.

lib/cloudant.js
```diff
@@ -948,7 +942,9 @@ Cloudant.prototype.updateIndex = function(mo, modelName, cb) {
     },
   };
   /* eslint-enable camelcase */
-  debug('Cloudant.prototype.updateIndex modelName %j', modelName);
+  var indexView = util.inspect(idx, 4);
+  debug('Cloudant.prototype.updateIndex modelName %s idx %s', modelName,
```
kjdelisle: Still didn't change the string formatting to be more readable:
`'Cloudant.prototype.updateIndex -- modelName: %s, idx: %s'`

test/automigrate.test.js
```diff
@@ -0,0 +1,30 @@
...
+    Foo = db.define('Foo', {
+      name: {type: String},
+    });
+    Bar = db.define('Bar', {
+      id: {type: Number, index: true},
+    });
+    db.automigrate(function verifyMigratedModel(err) {
+      if (err) return done(err);
+      Foo.create({name: 'foo'}, function(err, r) {
+        r.should.not.be.empty();
```
kjdelisle: No error check is being done here. What if verifyMigratedModel succeeds but Foo.create does not?

test/cloudant.test.js
```diff
@@ -104,7 +107,7 @@ describe('cloudant connector', function() {
 
   describe('replaceById', function() {
     after(function cleanUpData(done) {
-      Product.destroyAll(function(err) {
+      Product.destroyAll(function removeModelInstances(err) {
```
kjdelisle: Handle `err` please.

test/cloudant.test.js
```diff
@@ -47,16 +44,22 @@ describe('cloudant connector', function() {
...
+          done();
+        });
       });
     });
   });
 
   describe('replaceOrCreate', function() {
     after(function cleanUpData(done) {
-      Product.destroyAll(function(err) {
+      Product.destroyAll(function removeModelInstances(err) {
```
kjdelisle: Handle `err` please.

test/cloudant.test.js
```diff
@@ -47,16 +44,22 @@ describe('cloudant connector', function() {
...
-    Product.destroyAll(function(err) {
-      CustomerSimple.destroyAll(function(err) {
-        done();
+
+    db.automigrate(function cleanUpData(err) {
+      // automigrate only removes the design doc, but not instances' doc,
+      // so clean up data here just in case previous tests use same models.
+      if (err) return done(err);
+      Product.destroyAll(function removeModelInstances(err) {
+        CustomerSimple.destroyAll(function removeModelInstances(err) {
```
kjdelisle: Handle `err` please, both for the outer and inner calls.

test/automigrate.test.js
kjdelisle: I understand that if you get an `err` object back that `r` will most likely be undefined, but we want the actual error message in any test failures, rather than an assertion failure that hides the problem.

jannyHou:
@kjdelisle good catch, changed. PTAL again.

Issue #33, opened 2016-09-26
Enable juggler tests for  Cloudant/Couch CI


jannyHou:
Now we commented all test cases to get CI pass in https://github.com/strongloop/loopback-connector-cloudant/pull/29/files, but after CI config done we need to bring them back and make them pass all required platforms.

jannyHou:
Some update:
- Now all local test files are enabled on master branch: https://github.com/strongloop/loopback-connector-cloudant/commits/master

- Imported tests from juggler:
  - still undergoing
  - will look into every failure and either fix it(at least file bug) or document it as the gap between loopback connectors and cloudant(here I assume what juggler tests is our unified expected behaviour of a loopback connector)

jannyHou:
Failures:

- [ ] relations.test.js: [with filter skip](https://github.com/strongloop/loopback-datasource-juggler/blob/e9ff88f4536d0108712e802f89857bb170f393e1/test/relations.test.js#L734)

jannyHou:
Couchdb Test:

- Global config DOESNOT load when bddIf and bddDescribe get executed

- [x] basic-query.test.js
  - Fail [should support date "lt" that is satisfied](https://github.com/strongloop/loopback-datasource-juggler/blob/b79763d0652b833ca9dab709b194ce2d9a171b66/test/basic-querying.test.js#L390): shouldn't `null` less than any date?

jannyHou:
Close it since all PRs checked in.

PR #107, opened 2017-04-25
Fix sort query builder

jannyHou:
### Description
- [x] Bug

The sort query builder doesn't check first level property exist when it searches for its type.

### DIFF
--- a/test/cloudant.test.js
+++ b/test/cloudant.test.js
@@ -262,6 +262,8 @@ describe('cloudant connector', function() {
       });
       it('returns result when sorting type provided - missing first level' +
         'property', function(done) {
+        // Similar test case exist in juggler, but since it takes time to
+        // recover them, I temporarily add it here
         CustomerSimple.find({where: {'address.state': 'CA'},
           order: 'missingProperty:string'}, function(err, customers) {
           if (err) return done(err);
--- a/test/cloudant.test.js
+++ b/test/cloudant.test.js
@@ -260,17 +260,31 @@ describe('cloudant connector', function() {
             done();
           });
       });
-      it('returns result when sorting type provided', function(done) {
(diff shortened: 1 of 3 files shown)


jannyHou:
@slnode test please

test/cloudant.test.js
```diff
@@ -260,17 +260,33 @@ describe('cloudant connector', function() {
             done();
           });
       });
-      it('returns result when sorting type provided', function(done) {
+      it('returns result when sorting type provided - missing first level' +
+        'property', function(done) {
+        // Similar test case exist in juggler, but since it takes time to
+        // recover them, I temporarily add it here
```
loay: so what happens in the test case in juggler? Is it skipped?

test/maxrows.test.js
```diff
@@ -9,6 +9,9 @@ var should = require('should');
 var db, Thing;
 
 describe('cloudant max rows', function() {
+  // This test suite creates large number of data,
+  // require more time to complete data cleanUp
+  this.timeout(70000);
```
loay: Can't we add this to `package.json` instead?

test/cloudant.test.js
jannyHou: all test cases imported from juggler are skipped now, and I am recovering them: https://github.com/strongloop/loopback-connector-cloudant/issues/33

test/maxrows.test.js
jannyHou: @loay only this test case needs additional timeout, see my comments

jannyHou:
@loay FYI, this is the placeholder for recovering test cases in cloudant: https://github.com/strongloop/loopback-connector-cloudant/issues/33
and it's what I have been working on these few sprints :p some pr might connect to bug stories while those bugs are discovered during test cleanup.

PR #109, opened 2017-04-26
autoupdate and automigrate fix

jannyHou:
### Description

Expected behaviour:
`autoupdate`: updateIndex for models attached to db.
~`automigrate`: do autoupdate, then destroy model instances if model existing.~
`automigrate`: destroy model instances if model existing, then do destroy.

Current behaviour:
`autoupdate` and `automigrate` does same thing: just updateIndex

#### Related issues

- None

### Checklist

### DIFF
--- a/lib/cloudant.js
+++ b/lib/cloudant.js
@@ -969,7 +969,8 @@ Cloudant.prototype.automigrate = function(models, cb) {
 
 /**
   * Check if the models exist
-  * @param {String|String[]} [models] A model name or an array of model names. * If not present, apply to all models
+  * @param {String|String[]} [models] A model name or an array of model names.
+  * If not present, apply to all models
   * @param {Function} [cb] The callback function
   */
 
--- a/lib/cloudant.js
+++ b/lib/cloudant.js
@@ -926,7 +926,7 @@ Cloudant.prototype.autoupdate = function(models, cb) {
  * `ddoc` is the doc's _id, auto-generated as 'lb-index-ddoc' + modelName,
  * `name` is the index name, auto-generated as 'lb-index' + modelName.
  * It does NOT store properties in the doc.
- * `automigrate` one model destroys data if that model already exists.
+ * `automigrate` destroys model data if the model exists.
  * @param {String|String[]} [models] A model name or an array of model names.
  * If not present, apply to all models
  * @param {Function} [cb] The callback function
@@ -934,15 +934,31 @@ Cloudant.prototype.autoupdate = function(models, cb) {
 Cloudant.prototype.automigrate = function(models, cb) {
   debug('Cloudant.prototype.automigrate models %j', models);
   var self = this;
+  var existingModels = [];
   async.series([
+    function(callback) {
+      checkExistingModels(callback);
+    },
     function(callback) {
       destroyData(callback);
     },
     function(callback) {
       self.autoupdate(models, callback);
     }], cb);
+  function checkExistingModels(checkCb) {
+    async.eachSeries(models, function(model, cb1) {
+      self.isActual(model, function(err, exist) {
+        if (err) return cb1(err);
+        if (exist) existingModels.push(model);
+        cb1();
+      });
+    }, function(err) {
+      debug('Cloudant.prototype.automigrate checkExistingModels %j', err);
+      checkCb(err);
+    });
+  };
   function destroyData(destroyCb) {
-    async.eachSeries(models, function(model, cb2) {
+    async.eachSeries(existingModels, function(model, cb2) {
       self.destroyAll(model, {}, {}, cb2);
     }, function(err) {
       debug('Cloudant.prototype.automigrate %j', err);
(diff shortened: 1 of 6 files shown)


jannyHou:
caught by @ssh24 
> automigrate: do autoupdate, then destroy model instances if model existing.

That is wrong, should destroy old data first then autoupdate

lib/cloudant.js
```diff
@@ -879,41 +879,54 @@ Cloudant.prototype.selectModel = function(model, migrate) {
...
  * `name` is the index name, auto-generated as 'lb-index' + modelName.
- * It does NOT store properties in the doc, which makes autoupdate actually
- * does the same thing as automigrate.
- * It does NOT destroy previous model instances if model exists, user needs
- * to manually destroy them if necessary.
- *
- * @param {String|String[]} [models] A model name or an array of model names. If not
- * present, apply to all models
+ * It does NOT store properties in the doc.
+ * `automigrate` one model destroys data if that model already exists.
```
ssh24: `automigrate` destroys model data if the model exists.

test/automigrate.test.js
```diff
@@ -24,10 +24,23 @@ describe('cloudant automigrate', function() {
...
         r.name.should.equal('foo');
-        Foo.destroyAll(function(err) {
-          if (err) return done(err);
-          done();
-        });
+        done();
+      });
+    });
+  });
+  it('autoupdates models attache to db', function(done) {
```
ssh24: `autoupdate models attached to db`

test/cloudant.test.js
```diff
@@ -45,18 +45,7 @@ describe('cloudant connector', function() {
...
-      if (err) return done(err);
-      Product.destroyAll(function removeModelInstances(err) {
-        if (err) return done(err);
-        CustomerSimple.destroyAll(function removeModelInstances(err) {
-          if (err) return done(err);
-          done();
-        });
-      });
-    });
+    db.automigrate(done);
```
ssh24: Could we check after doing `automigrate` that the data was destroyed?

lib/cloudant.js
```diff
@@ -879,41 +879,54 @@ Cloudant.prototype.selectModel = function(model, migrate) {
...
  * @param {Function} [cb] The callback function
  */
 Cloudant.prototype.autoupdate = function(models, cb) {
   debug('Cloudant.prototype.autoupdate %j', models);
-  this.automigrate(models, cb);
+  var self = this;
+  async.eachSeries(models, function(model, cb2) {
+    debug('Cloudant.prototype.autoupdate model %j', model);
+    var mo = self.selectModel(model, true);
+    self.updateIndex(mo, model, cb2);
```
b-admike: maybe I am wrong, but do you mean `cb` here instead of `cb2`?

jannyHou: @b-admike I think here it should be `cb2`, it is the callback function of `async.eachSeries`'s iteration function.
The function next line is the "finish" function after all iterations done, in this function, we `cb(err)`, so err will be returned through `cb` if exist.

test/cloudant.test.js
jannyHou: Good point, created a test case for it in https://github.com/strongloop/loopback-connector-cloudant/pull/109/files#diff-d4318f238579a6d222dac940aab8b69dR50

lib/cloudant.js
kjdelisle: Please don't do this, as it's not necessary. The innermost definition of cb will be the one that gets used, so there's no need to differentiate between them unless you have special logic where you might call one of the callbacks, but not the other.

If that was the case, then I would definitely want a better naming convention than `cb2`. :/

lib/cloudant.js
```diff
@@ -879,43 +879,111 @@ Cloudant.prototype.selectModel = function(model, migrate) {
...
+      checkExistingModels(callback);
+    },
+    function(callback) {
+      destroyData(callback);
+    },
+    function(callback) {
+      self.autoupdate(models, callback);
+    }], cb);
+  function checkExistingModels(checkCb) {
+    async.eachSeries(models, function(model, cb1) {
```
kjdelisle: Change `cb1` please.

lib/cloudant.js
```diff
@@ -879,43 +879,111 @@ Cloudant.prototype.selectModel = function(model, migrate) {
...
+      self.isActual(model, function(err, exist) {
+        if (err) return cb1(err);
```
kjdelisle: fix

lib/cloudant.js
```diff
@@ -879,43 +879,111 @@ Cloudant.prototype.selectModel = function(model, migrate) {
...
+        if (exist) existingModels.push(model);
+        cb1();
+      });
+    }, function(err) {
+      debug('Cloudant.prototype.automigrate checkExistingModels %j', err);
+      checkCb(err);
+    });
+  };
+  function destroyData(destroyCb) {
+    async.eachSeries(existingModels, function(model, cb2) {
```
kjdelisle: Change `cb2`.

lib/cloudant.js
```diff
@@ -879,43 +879,111 @@ Cloudant.prototype.selectModel = function(model, migrate) {
...
+      self.destroyAll(model, {}, {}, cb2);
```
kjdelisle: fix

jannyHou:
@slnode test please

lib/cloudant.js
```diff
@@ -901,43 +901,116 @@ Cloudant.prototype.selectModel = function(model, migrate) {
...
+    }, function(err) {
+      debug('Cloudant.prototype.automigrate %j', err);
+      destroyCb(err);
+    });
+  };
+};
+
+/**
+  * Check if the models exist
+  * @param {String|String[]} [models] A model name or an array of model names. * If not present, apply to all models
```
0candy: Was the second part of this sentence suppose to be on a new line? I see an asterisk.

jannyHou: good catch, vs code shows them in separate lines, fixed.

jannyHou:
@slnode test please

jannyHou:
@0candy Comment applied, PTAL again, thanks.

Issue #31, opened 2016-09-20
Cloudant - Model property with type number and marked ID expects string as value


cgole:
PMR link: https://jazzop27.rtp.raleigh.ibm.com:9443/ccm/web/projects/CloudOE#action=com.ibm.team.workitem.viewWorkItem&id=210887

## Problem:
Attempting to insert a record into Cloudant with a numeric ID will fail, since the inserted object does not have an id of type String.

## Potential Fix
Investigate whether or not calling `.toString()` on the `_id` parameter before an insert request is made will fix the issue.

## Acceptance Criteria
- Numeric IDs should be valid for inserting records into Cloudant
- Test case(s) should validate this behaviour
- Document this workaround on loopback-connector-cloudant

cgole:
Since cloudant uses documents to store data then we can only use type `string` for id:
see https://docs.cloudant.com/document.html

jannyHou:
Add some details from discussion:

```json
"id": {
     "type": "number",
     "id": true,
     "required": true,
     "generated": false
}
```
Will see error 
> Document id must be a string

I am figuring out how loopback converts the id type to string, will update details later. Not sure which one is better, giving a warning or throwing an error, but we should add some validation/check instead of silently changing it to another type.

cgole:
@deepakrkris this needs to be fixed asap. Customer is pushing for a fix

ssh24:
Issue fixed on #115. Moving it to TBR.

PR #115, opened 2017-05-02
Allow id property to be a number

ssh24:
### Description
connect to https://github.com/strongloop/loopback-connector-cloudant/issues/31

#### Related issues

- None

### Checklist

### DIFF
--- a/lib/cloudant.js
+++ b/lib/cloudant.js
@@ -108,7 +108,8 @@ Cloudant.prototype.connect = function(cb) {
  * @param {Object} doc The model document/data
  */
 Cloudant.prototype.toDB = function(modelName, modelObject, doc) {
-  var idValue = this.getIdValue(modelName, doc);
+  var idValue = this.getIdValue(modelName, doc) &&
+     this.getIdValue(modelName, doc).toString();
   var idName = this.idName(modelName);
   if (!doc) {
     doc = {};
@@ -292,6 +293,25 @@ function findRecursive(mo, query, docs, include, cb) {
   mo.db.find(query, function(err, rst) {
     debug('Cloudant.prototype.all (findRecursive) results: %j %j', err, rst);
     if (err) return cb(err);
+
+    // sort back the ids in "numerical order" if they are numeric
+    query.sort.forEach(function(queries) {
+      if (queries.hasOwnProperty('_id:number')) {
+        var sortType = queries['_id:number'];
+        if (Array.isArray(rst.docs)) {
+          if (sortType === 'desc')
+            rst.docs.sort(function(a, b) {
+              return parseInt(a._id) - parseInt(b._id);
+            }).reverse();
+          else {
+            rst.docs.sort(function(a, b) {
+              return parseInt(a._id) - parseInt(b._id);
+            });
+          }
+        }
+      }
+    });
+
     // work around for issue
     // https://github.com/strongloop/loopback-connector-cloudant/issues/73
     if (!rst.docs) {
@@ -350,6 +370,8 @@ Cloudant.prototype.buildSelector = function(model, mo, where) {
     }
     if (k === idName) {
       k = '_id';
+      cond = (typeof cond === 'object' || Array.isArray(cond)) ? cond :
+        cond.toString();
     }
     var spec = false;
     var options = null;
@@ -545,13 +567,12 @@ Cloudant.prototype.buildSort = function(mo, model, order) {
       fieldType = props[n] && props[n].type;
     }
 
+    if (n === idName)
+      n = '_id';
     if (fieldType === Number)
       n = n.concat(':number');
     if (isString(fieldType) || isDate(fieldType))
       n = n.concat(':string');
-    if (n === idName) {
-      n = '_id';
-    }
     function isString(type) {
       if (type === String || type === 'string' || type === 'String')
         return true;
--- a/lib/cloudant.js
+++ b/lib/cloudant.js
@@ -108,8 +108,9 @@ Cloudant.prototype.connect = function(cb) {
  * @param {Object} doc The model document/data
  */
 Cloudant.prototype.toDB = function(modelName, modelObject, doc) {
-  var idValue = this.getIdValue(modelName, doc) &&
-     this.getIdValue(modelName, doc).toString();
+  // toString() this value because IDs must be strings: https://docs.cloudant.com/document.html
+  var idValue = this.getIdValue(modelName, doc);
+  if (idValue) idValue = idValue.toString();
   var idName = this.idName(modelName);
   if (!doc) {
     doc = {};
@@ -268,7 +269,7 @@ Cloudant.prototype.all = function all(model, filter, options, cb) {
   /* eslint-enable camelcase */
   if (filter.offset) query.skip = filter.offset;
   if (filter.limit) query.limit = filter.limit;
-  query.sort = self.buildSort(mo, model, filter.order);
+  if (filter.order) query.sort = self.buildSort(mo, model, filter.order);
   debug('Cloudant.prototype.all %j %j %j', model, filter, query);
   include = function(docs, cb) {
     if (!options || !options.raw) {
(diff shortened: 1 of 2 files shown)


ssh24:
@slnode test please

ssh24:
@slnode test please

test/cloudant.test.js
```diff
@@ -295,6 +310,124 @@ describe('cloudant connector', function() {
...
+            should.equal(result.length, 3);
+            should.deepEqual(result[0].__data, data[0]);
+            should.deepEqual(result[1].__data, data[1]);
+            should.deepEqual(result[2].__data, data[2]);
+            done();
+          });
+        });
+    });
+
+    it('destory instances with numerical id (destroyById)', function(done) {
```
0candy: typo `destroy`

test/cloudant.test.js
```diff
@@ -295,6 +310,124 @@ describe('cloudant connector', function() {
...
+          should.not.exist(err);
+          should.equal(result.length, 2);
+          should.deepEqual(result[0].__data, data[0]);
+          should.deepEqual(result[1].__data, data[2]);
+          done();
+        });
+      });
+    });
+
+    it('destory instances with "where" filter', function(done) {
```
0candy: typo `destroy`

test/cloudant.test.js
ssh24: Fixed :)

test/cloudant.test.js
```diff
@@ -295,6 +310,124 @@ describe('cloudant connector', function() {
...
+        should.not.exist(err);
+        should.equal(result.length, 1);
+        should.deepEqual(result[0].__data, data[0]);
+        done();
+      });
+    });
+
+    it('find instances with "order" filter (ASC)', function(done) {
+      SimpleEmployee.find({order: 'id ASC'}, function(err, result) {
+        should.not.exist(err);
```
b-admike: I suggest `if (err) done(err)` here instead of `should.not.exist(err)` so that it doesn't run the other checks past this point.

test/cloudant.test.js
```diff
@@ -295,6 +310,124 @@ describe('cloudant connector', function() {
...
+        should(result[0].id).equal(data[0].id);
+        should(result[1].id).equal(data[1].id);
+        should(result[2].id).equal(data[2].id);
+        done();
+      });
+    });
+
+    it('find instances with "order" filter (DESC)', function(done) {
+      SimpleEmployee.find({order: 'id DESC'}, function(err, result) {
+        should.not.exist(err);
```
b-admike: `if (err) done(err)`

test/cloudant.test.js
```diff
@@ -295,6 +310,124 @@ describe('cloudant connector', function() {
...
+        id: data[1].id,
+        name: 'Christian Thompson',
+        age: 32,
+      };
+      data[1].name = 'Christian Thompson';
+      data[1].age = 32;
+
+      SimpleEmployee.replaceById(data[1].id, updatedData,
+        function(err, result) {
+          should.not.exist(err);
```
b-admike: `if (err) done (err)`

test/cloudant.test.js
```diff
@@ -295,6 +310,124 @@ describe('cloudant connector', function() {
...
+          should.equal(result.id, data[1].id);
+          should.equal(result.name, updatedData.name);
+          should.equal(result.age, updatedData.age);
+
+          SimpleEmployee.find(function(err, result) {
+            should.not.exist(err);
```
b-admike: same as above

test/cloudant.test.js
```diff
@@ -295,6 +310,124 @@ describe('cloudant connector', function() {
...
+            should.deepEqual(result[1].__data, data[1]);
+            should.deepEqual(result[2].__data, data[2]);
+            done();
+          });
+        });
+    });
+
+    it('destroy instances with numerical id (destroyById)', function(done) {
+      SimpleEmployee.destroyById(data[1].id, function(err, result) {
+        should.not.exist(err);
```
b-admike: same as above

test/cloudant.test.js
```diff
@@ -295,6 +310,124 @@ describe('cloudant connector', function() {
...
+          should.deepEqual(result[0].__data, data[0]);
+          should.deepEqual(result[1].__data, data[2]);
+          done();
+        });
+      });
+    });
+
+    it('destroy instances with "where" filter', function(done) {
+      SimpleEmployee.destroyAll({id: data[2].id}, function(err, result) {
+        should.not.exist(err);
```
b-admike: same as above

test/cloudant.test.js
```diff
@@ -295,6 +310,124 @@ describe('cloudant connector', function() {
...
+        should(result).have.property('count');
+        should.equal(result.count, 1);
+
+        SimpleEmployee.find(function(err, result) {
+          should.not.exist(err);
```
b-admike: same as above

test/cloudant.test.js
ssh24: I do not think that is correct. The way `should` works, if it fails on the first check it fails the whole test case. So if it fails on `should.not.exist(err);` it will not check the other assertions.

lib/cloudant.js
```diff
@@ -108,7 +108,8 @@ Cloudant.prototype.connect = function(cb) {
  * @param {Object} doc The model document/data
  */
 Cloudant.prototype.toDB = function(modelName, modelObject, doc) {
-  var idValue = this.getIdValue(modelName, doc);
+  var idValue = this.getIdValue(modelName, doc) &&
+     this.getIdValue(modelName, doc).toString();
```
kjdelisle: Let's not do this; it requires two calls to the same function, which means you're actually doing all that work twice!

```
// We toString this value because IDs must be strings: <put link to Cloudant API doc here>
var idValue = this.getIdValue(modelName, doc);
if (!idValue) { // do error stuff here } else { idValue = idValue.toString(); }
// etc
```

lib/cloudant.js
```diff
@@ -292,6 +293,25 @@ function findRecursive(mo, query, docs, include, cb) {
   mo.db.find(query, function(err, rst) {
     debug('Cloudant.prototype.all (findRecursive) results: %j %j', err, rst);
     if (err) return cb(err);
+
+    // sort back the ids in "numerical order" if they are numeric
+    query.sort.forEach(function(queries) {
```
kjdelisle: Why are we sorting the list for them? What if they don't want the list sorted? Now they have no choice!

test/cloudant.test.js
```diff
@@ -295,6 +310,124 @@ describe('cloudant connector', function() {
...
+    before(function(done) {
+      SimpleEmployee.create(data, function(err, result) {
+        if (err) done(err);
+        done(null, result);
+      });
+    });
+
+    it('find instances with numeric id (findById)', function(done) {
+      SimpleEmployee.findById(data[1].id, function(err, result) {
+        should.not.exist(err);
```
kjdelisle: `if (err) return done(err)`

test/cloudant.test.js
kjdelisle: `if (err) return done(err);`
Don't want it to fall through afterwards!

test/cloudant.test.js
kjdelisle: `if (err) return done(err);`

test/cloudant.test.js
kjdelisle: `if (err) return done(err);`

test/cloudant.test.js
kjdelisle: `if (err) return done(err);`

test/cloudant.test.js
kjdelisle: `if (err) return done(err);`

test/cloudant.test.js
kjdelisle: `if (err) return done(err);`

lib/cloudant.js
ssh24: Applied feedback.

kjdelisle:
Sorry, @ssh24  just noticed your reply to @b-admike regarding the mocha behaviour for failures. If that's definitely handling the done callback for us, then that's fine.

lib/cloudant.js
ssh24: We only sort them if they give want to sort them in order. i.e if the end user filters it by 
`{order : { id ASC}}` .

test/cloudant.test.js
```diff
@@ -295,6 +310,124 @@ describe('cloudant connector', function() {
...
+    }, {
+      id: 12,
+      name: 'Michael Santer',
+      age: 30,
+    }];
+
+    before(function(done) {
+      SimpleEmployee.create(data, function(err, result) {
+        if (err) done(err);
+        done(null, result);
```
jannyHou: hmm, any reason why callback `result` here? `SimpleEmployee.create(data,done);` would be enough IMO.

test/cloudant.test.js
jannyHou: @ssh24 `should.not.exist(err)` and `if (err) return done(err)` does same check, and both skip other assertions if error is not null, the difference is the later one gives you a better error log.

When mocha handles the error, it consoles a well formatted error message, while `should` will display it as a failed assertion, expected, actual, many lines of +, -, etc...

test/cloudant.test.js
```diff
@@ -295,6 +310,124 @@ describe('cloudant connector', function() {
...
+        should.equal(result.length, 1);
+        should.deepEqual(result[0].__data, data[0]);
+        done();
+      });
+    });
+
+    it('find instances with "order" filter (ASC)', function(done) {
+      SimpleEmployee.find({order: 'id ASC'}, function(err, result) {
+        should.not.exist(err);
+        should(result[0].id).equal(data[0].id);
```
jannyHou: better to add a `result.length.should.equal(3)` before checking `result` details. Not necessary, but think of if result is null or missing data, that check tells you what's going wrong faster than reading 3 failed assertions and figure out the reason :p

test/cloudant.test.js
ssh24: Ah, right. Fixed :)

test/cloudant.test.js
ssh24: Ah, right. Fixed :)

ssh24:
@slnode test please

ssh24:
@slnode test please

Issue #47, opened 2016-10-19
updateAttributes doesn't replace the property with array type.


jannyHou:
#Same problem as `updateAll`: https://github.com/strongloop/loopback-connector-cloudant/issues/44

Reproduce:

A model instance with an array type property

``` js
  modelInstance: {
     name: 'test',
     arrProperty: [
       {id: 1, name: original1},
       {id: 2, name: original2},
       {id: 3, name: original3}
     ] 
  }
```

When do updateAttributes:

``` js
modelInstance.updateAttributes({arrProperty: [
  {id: 1, name: new1},
]})
```

The expected new arrProperty is 

``` js
arrProperty: [{id: 1, name: new1}]
```

But actual result is

``` js
arrProperty: [
       {id: 1, name: new1},
       {id: 2, name: original2},
       {id: 3, name: original3}
     ] 
```

rimvydas-urbonas:
What is the status of this issue?

jannyHou:
@rimvydas-urbonas We label the issue as "Needs Priority" but it's not in backlog yet :(

rimvydas-urbonas:
Maybe you can suggest workaround to update array property by replacing value not appending?
Thanks in advance.

kjdelisle:
Being unable to update array properties seems fairly major to me. Going to add this to our backlog.

gunjpan:
@jannyHou : The root cause for this issue is `_.merge()` at [lib/cloudant.js : L517](https://github.com/strongloop/loopback-connector-cloudant/blob/35580d0df5ce322f194e3c5b606e4230965ae021/lib/cloudant.js#L517). This should be replaced with `_.mergeWith(object, sources, customizer)` where `customizer` is the function to handle Arrays. See more [here](https://lodash.com/docs/#mergeWith)

jannyHou:
I think it's already fixed in https://github.com/strongloop/loopback-connector-cloudant/pull/117, closing it. Feel free to reopen it if you still run into same error.

Issue #44, opened 2016-10-18
updateAll doesn't replace the property with array type. 


jannyHou:
Reproduce:
1. Create a model `Test`
2. Create an instance for `Test`:
   
   ``` js
   testInstance: {
     id: 1,
     name: 'test1'
     skills: [
       {id: 0, text: "JAVA"},
       {id: 1, text: "CSS"},
       {id: 2, text: "HTML" },
       {id: 3, text: "JAVASCRIPT"},
       {id: 4, text: ".NET"}
     ]
   }
   ```
3. Try updateAll:
   
   ``` js
   Test.updateAll({id: test.id}, {skills: [
   { id: 1, text: "newJAVA"},
   { id: 2, text: "newCSS"}
   ]}, function(err, count){} );
   ```
4. The updated skills are expected to be:
   
   ``` js
    skills: [
      { id: 1, text: "newJAVA"},
      { id: 2, text: "newCSS"}
    ]
   ```
   
   But actual result is:
   
   ``` js
   skills: [
     {id: 0, text: "newJAVA"},
     {id: 1, text: "newCSS"},
     {id: 2, text: "HTML" },
     {id: 3, text: "JAVASCRIPT"},
     {id: 4, text: ".NET"}
   ]
   ```

Details see: http://stackoverflow.com/questions/40111147/updateall-loopback-datasource-does-not-delete-deletions-inside-updated-data

gunjpan:
@jannyHou : Same issue as: https://github.com/strongloop/loopback-connector-cloudant/issues/47#issuecomment-274858304

Root cause: https://github.com/strongloop/loopback-connector-cloudant/blob/35580d0df5ce322f194e3c5b606e4230965ae021/lib/cloudant.js#L574

jannyHou:
I think it's already fixed in https://github.com/strongloop/loopback-connector-cloudant/pull/117, closing it. Feel free to reopen it if you still run into same error.

PR #117, opened 2017-05-03
overwrite array prop value on update

b-admike:
### Description

Connected to https://github.com/strongloop-internal/scrum-apex/issues/62
#### Related issues

#47 
#44 

### Checklist

### DIFF
--- a/lib/cloudant.js
+++ b/lib/cloudant.js
@@ -705,14 +705,16 @@ Cloudant.prototype.updateAttributes = function(model, id, data, options, cb) {
   debug('Cloudant.prototype.updateAttributes %j %j %j',
         model, id, data, options);
   var self = this;
-  var mo = self.selectModel(model);
-  mo.db.get(id, function(err, doc) {
-    if (err) return cb(err, doc);
-
-    data = getPlainJSONData.call(self, model, data);
-    _.merge(doc, data);
-    self.create(model, doc, options, function(err) {
-      cb(err, self.fromDB(model, mo, doc));
+  if (!id) {
+    var idName = self.idName(model);
+    id = data[idName];
+  }
+  self.getCurrentRevision(model, id, function(err, rev) {
+    if (err) return cb(err);
+    if (rev) data._rev = rev;
+    self.create(model, data, options, function(err) {
+      if (err) return cb(err);
+      return cb(err, data);
     });
   });
 };
@@ -779,16 +781,16 @@ Cloudant.prototype.updateAll = function updateAll(model, where,
     if (docs.length === 0) {
       return cb(null, {count: 0});
     }
-
-    data = getPlainJSONData.call(self, model, data);
-
-    for (var i = 0; i < docs.length; i++) {
-      _.merge(docs[i], data);
-    }
-    debug('Cloudant.prototype.updateAll bulk docs: %j', docs);
-    mo.db.bulk({docs: docs}, function(err, result) {
-      if (result) result.count = docs.length;
-      cb(err, result);
+    async.each(docs, function(doc, cb) {
+      self.updateAttributes(doc.loopback__model__name, null, data, options, cb);
+    }, function(err) {
+      if (err) return cb(err);
+      debug('Cloudant.prototype.updateAll bulk docs: %j', docs);
+      mo.db.bulk({docs: docs}, function(err, result) {
+        if (err) return cb(err);
+        if (result) result.count = docs.length;
+        cb(err, result);
+      });
     });
   });
 };
--- a/lib/cloudant.js
+++ b/lib/cloudant.js
@@ -705,10 +705,15 @@ Cloudant.prototype.updateAttributes = function(model, id, data, options, cb) {
   debug('Cloudant.prototype.updateAttributes %j %j %j',
         model, id, data, options);
   var self = this;
-  data._id = id;
-  self.create(model, data, options, function(err) {
+  var mo = self.selectModel(model);
+  mo.db.get(id, function(err, doc) {
     if (err) return cb(err);
-    return cb(err, data);
+    data = getPlainJSONData.call(self, model, data);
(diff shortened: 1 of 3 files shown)


test/cloudant.test.js
```diff
@@ -197,6 +200,93 @@ describe('cloudant connector', function() {
...
+    });
+
+    after(function(done) {
+      Product.destroyAll(function(err) {
+        if (err) return done(err);
+        done();
+      });
+    });
+
+    it('updates a single model with array props',
```
jannyHou: nitpick: `updates a single instance with array props`, `instance` is clearer than `model`

test/cloudant.test.js
```diff
@@ -197,6 +200,93 @@ describe('cloudant connector', function() {
...
+              res.__data.releases.should.deepEqual(prod2.releases);
+              res.__data.type.should.deepEqual(prod2.type);
+              res.__data.foo.should.deepEqual(prod2.foo);
+              done();
+            });
+          });
+        });
+      });
+
+    it('updates all models with array props',
```
jannyHou: same here: 'updates all instances with array props'

test/cloudant.test.js
```diff
@@ -197,6 +200,93 @@ describe('cloudant connector', function() {
...
+        done();
+      });
+    });
+
+    it('updates a single model with array props',
+      function(done) {
+        prod1.setAttribute('type', ['cinnamon raisin']);
+        prod1.setAttribute('releases', [4, 5, 6]);
+        prod1.setAttribute('foo', [{id: 3, name: 'bread3'}]);
+        Product.updateAll({id: 1}, prod1, {}, function(err, res) {
```
ssh24: I think just the callback `function(err, res)` should be good. Do not need specify the empty `options` object because the function itself should do it. If it does not do that, we need to fix that first.

test/cloudant.test.js
```diff
@@ -197,6 +200,93 @@ describe('cloudant connector', function() {
...
+          if (err) return done(err);
```
ssh24: Can we do `should.not.exist(err)` here instead?

test/cloudant.test.js
```diff
@@ -197,6 +200,93 @@ describe('cloudant connector', function() {
...
+     var data = {
+       price: 200,
+       releases: [7],
+       type: ['everything'],
+       foo: [{id: 1, name: 'bar'}],
+     };
+
+     Product.updateAll({price: 100}, data, {}, function(err, res) {
+       should.not.exist(err);
+       Product.find({}, function(err, res) {
```
ssh24: Same here. Do not need to specify an empty filter object. `Product.find(function(err,res)` is sufficient.

test/cloudant.test.js
```diff
@@ -197,6 +200,93 @@ describe('cloudant connector', function() {
...
+
+    it('updates a single model with array props',
+      function(done) {
+        prod1.setAttribute('type', ['cinnamon raisin']);
+        prod1.setAttribute('releases', [4, 5, 6]);
+        prod1.setAttribute('foo', [{id: 3, name: 'bread3'}]);
+        Product.updateAll({id: 1}, prod1, {}, function(err, res) {
+          if (err) return done(err);
+          Product.findById('1', function(err, res) {
+            should.not.exist(err);
```
jannyHou: - I think we might need more assertions to verify it's a full update(the original purpose of this issue)
- And could you also verify that if we update a specific field, the new data will override the old one? 
```js
prod1.setAttribute('foo', [{id: 1, name: 'newbread1'}, {id: 3, name: 'bread3'}]);
```
```js
var result = res._data;

result.releases.length.should.equal(3);
result. releases[0].should.equal(4);
result. releases[1].should.equal(5);
result. releases[2].should.equal(6);

result.type.length.should.equal(1); 
// this is what I am interested to know, 
// will cloduant reset other fields to null or totally delete them?
result.type[0].should.equal('cinnamon raisin');

result.foo.length.should.equal(2);
result.foo[the_index_of_id_3].name.should.equal('bread3');
result.foo[the_index_of_id_1].name.should.equal('newbread1');
```

lib/cloudant.js
```diff
@@ -773,22 +768,16 @@ Cloudant.prototype.updateAll = function updateAll(model, where,
...
-    data = getPlainJSONData.call(self, model, data);
-
-    for (var i = 0; i < docs.length; i++) {
-      _.merge(docs[i], data);
-    }
-    debug('Cloudant.prototype.updateAll bulk docs: %j', docs);
-    mo.db.bulk({docs: docs}, function(err, result) {
-      if (result) result.count = docs.length;
-      cb(err, result);
+    async.each(docs, function(doc, cb) {
```
jannyHou: hmm, this solution looks good, while my concern is here we have to sacrifice the time we benefit from only talking to db once by `bulk`

test/cloudant.test.js
jannyHou: @ssh24 could you take a look of https://github.com/strongloop/loopback-connector-cloudant/pull/115/files#r114569262 :)
Actually I think `if (err) return done(err)` would be better... IIRC, we spent huge effort in juggler to refactor the assertion from using should to using return done(err) :p

Copied from that comment:
> should.not.exist(err) and if (err) return done(err) does same check, and both skip other assertions if error is not null, the difference is the later one gives you a better error log.
When mocha handles the error, it consoles a well formatted error message, while should will display it as a failed assertion, expected, actual, many lines of +, -, etc...

ssh24: @jannyHou I did read your comment as it was in my PR and agree that it does the same. However, it is not consistent in this PR. 2 lines down it uses `should.not.exist(err);`. So we should make them consistent on whichever way we are using. :)

lib/cloudant.js
b-admike: changed back to `bulk` :-)

PR #111, opened 2017-04-26
Fix for updateOrCreate (upsert) when id is supplied

djorg83:
### Description
Fix for upsert when supplying an id.

#### Related issues

- #110 

### Checklist

- [x ] New tests added or existing tests modified to cover all changes
- [x ] Code conforms with the [style
  guide](http://loopback.io/doc/en/contrib/style-guide.html)

PR #121, opened 2017-05-05
Fix/upsert error

jannyHou:
related to https://github.com/strongloop/loopback-connector-cloudant/pull/111

### DIFF
--- a/lib/cloudant.js
+++ b/lib/cloudant.js
@@ -13,7 +13,7 @@ var debug = require('debug')('loopback:connector:cloudant');
 var async = require('async');
 var url = require('url');
 var util = require('util');
-var _ = require('lodash');
+var _ = require('lodash'); //
 
 /**
  * Initialize the Cloudant connector for the given data source
@@ -745,7 +745,9 @@ Cloudant.prototype.updateOrCreate = function(model, data, cb) {
   var id = data[idName];
   if (id) {
     self.getCurrentRevision(model, id, function(err, rev) {
-      if (err) return cb(err);
+      if (err && err.statusCode !== 404) {
+        return cb(err);
+      }
       if (rev) data._rev = rev;
       self.create(model, data, {}, function(err) {
         if (err) return cb(err);
--- a/lib/cloudant.js
+++ b/lib/cloudant.js
@@ -13,7 +13,7 @@ var debug = require('debug')('loopback:connector:cloudant');
 var async = require('async');
 var url = require('url');
 var util = require('util');
-var _ = require('lodash'); //
+var _ = require('lodash');
 
 /**
  * Initialize the Cloudant connector for the given data source
--- a/test/cloudant.test.js
+++ b/test/cloudant.test.js
@@ -291,7 +291,24 @@ describe('cloudant connector', function() {
      });
    });
   });
+  describe('updateOrCreate', function() {
+    afterEach(function(done) {
+      Product.destroyById(1, function(err) {
+        if (err) return done(err);
+        done();
+      });
+    });
 
+    it('can call upsert with an id and have it create a document without error',
+      function(done) {
+        Product.upsert({
+          id: '1'
+        }, function(err, res) {
+          if (err) return done(err);
+          done();
+        });
+      });
+  });
   // the test suite is to make sure when
   // user queries against a non existing property
   // the app won't crash
--- a/test/cloudant.test.js
+++ b/test/cloudant.test.js
@@ -301,11 +301,23 @@ describe('cloudant connector', function() {
 
     it('can call upsert with an id and have it create a document without error',
       function(done) {
+        // scenario: upsert does a create
         Product.upsert({
-          id: '1'
+          id: '1',
+          name: 'bread',
         }, function(err, res) {
           if (err) return done(err);
-          done();
+          res.id.should.equal('1');
+          res.name.should.equal('bread');
+          // scenario: upsert does a update
+          Product.upsert({
+            id: '1',
+            name: 'newBread',
+          }, function(err, res) {
+            res.id.should.equal('1');
+            res.name.should.equal('newBread');
+            done();
+          });
         });
       });
   });


jannyHou:
@djorg83 I open a new PR for you, regarding https://github.com/strongloop/loopback-connector-cloudant/pull/111#issuecomment-299503072
Are you ok if I check in this one?

lib/cloudant.js
```diff
@@ -745,7 +745,9 @@ Cloudant.prototype.updateOrCreate = function(model, data, cb) {
   var id = data[idName];
   if (id) {
     self.getCurrentRevision(model, id, function(err, rev) {
-      if (err) return cb(err);
+      if (err && err.statusCode !== 404) {
```
ssh24: @jannyHou What does `statusCode` return on getting the current revision? Both on happy and unhappy cases.

test/cloudant.test.js
```diff
@@ -291,7 +291,24 @@ describe('cloudant connector', function() {
...
+        if (err) return done(err);
+        done();
+      });
+    });
 
+    it('can call upsert with an id and have it create a document without error',
+      function(done) {
+        Product.upsert({
+          id: '1'
+        }, function(err, res) {
```
ssh24: Can we assert more on the `res`?

lib/cloudant.js
kjdelisle: If the record exists, you'll get it (and a 200 OK). If not, you'll get a 404.

If you get other errors like a document conflict (409), then it'll callback.

lib/cloudant.js
```diff
@@ -745,7 +745,9 @@ Cloudant.prototype.updateOrCreate = function(model, data, cb) {
...
+        return cb(err);
+      }
```
kjdelisle: What happens if I do this and I end up getting 200 OK?
The record exists, and now we're calling create on top of it.

Should the body of this function not be something like this?
```
self.updateAttributes(model, id, data, options, function(err, result) {
  if (err) {
     if (err.statusCode === 404) {
        // create logic here
     } else {
        return cb(err);
     }
  } else {
     // do whatever else you need to do, then return results.
  }
}
```

lib/cloudant.js
jannyHou: @ssh24 Please check https://github.com/djorg83/loopback-connector-cloudant/blob/5c96f2352edd2c10fa7ed8399da38c5a12302064/lib/cloudant.js#L241-L245

ssh24: Got it. Thanks.

lib/cloudant.js
jannyHou: @kjdelisle here function `create` is actually doing update, when you check the code, here `create` will do a `_insert` with `id` and `_rev` included.

jannyHou: https://github.com/cloudant/nodejs-cloudant/blob/master/example/crud.js#L50 `insert` does both create and update

jannyHou:
@kjdelisle currently we could not do "record exists, but rev does not match (should give 409)", we don't have control of `rev` property

test/cloudant.test.js
```diff
@@ -291,7 +291,7 @@ describe('cloudant connector', function() {
      });
    });
   });
-  describe('updateOrCreate', function() {
+  describe.only('updateOrCreate', function() {
```
ssh24: You should remove `.only` :)

jannyHou: ah, sorry, was surprised why this time CI pass so fast

lib/cloudant.js
kjdelisle: I understand, but if we want to preserve the idea of what `updateOrCreate` does from a LoopBack perspective, then we should use `updateAttributes` to attempt the update, and then handle the failure as appropriate, either by making the record, or returning the error up the callback chain.

kjdelisle: >https://github.com/cloudant/nodejs-cloudant/blob/master/example/crud.js#L50 insert does both create and update

`nodejs-cloudant` shouldn't be determining our behaviour here; update from a LB perspective is not an alias for replace/create. I know that previously, I said that we should make it clear to users that we'll just do what the driver does, but now I'm not so sure.

What are the potential pitfalls of using `updateAttributes` as the default action here?

jannyHou: @kjdelisle hmm, I don't get this idea: "but if we want to preserve the idea of what updateOrCreate does from a LoopBack perspective, then we should use updateAttributes to attempt the update", from my understanding, from a BL perspective it's the way inverse, `updateAttributes` should call `updateAll`(alias for update).
And if updateOrCreate calls other method, it should be `update` but not `updateAttributes`

kjdelisle: `updateOrCreate` seems to strongly imply that it'll update the record (not in the Cloudant sense!) or create it. `updateAttributes` is the only "true" update we have in that sense, is it not?

jannyHou: @kjdelisle 
> What are the potential pitfalls of using updateAttributes as the default action here

I could not think of how to deal with `updateAttributes` here lol, by `updateOrCreate`, you give cloudant a data without knowing is it exist or not, neither what does the data look in it.
While by 'updateAttributes', you know exactly which properties you want to update. Calling updateAttributes we have to fetch original data then magically "merge" it with our data, then update certain properties...

kjdelisle:
>currently in tests, we could not do "record exists, but rev does not match (should give 409)", we don't have control of rev property

We might want to land the `_.rev` PR first, then. We can chat about it when I'm back online. :)

lib/cloudant.js
jannyHou: @kjdelisle there is another method called `update/updateAll`, I think you are actually looking for that: https://github.com/strongloop/loopback-connector-cloudant/blob/347a75289abaeb9040f64e43195144d68e8c711d/lib/cloudant.js#L770

jannyHou:
@kjdelisle How about I create this PR against 1.x branch? I feel if we can include '_rev', I don't have to waste so much time on making it work on master(2.x) but finally just reject it in favour of `_rev` PR :( Take your time let's discuss afterwards.

jannyHou:
close in favor of https://github.com/strongloop/loopback-connector-cloudant/pull/129

Issue #55, opened 2016-11-14
Keep ._rev property on returned instance of model


jannyHou:
A further discussion of https://github.com/strongloop/loopback-connector-cloudant/pull/54#issuecomment-259401952
Regarding
> When I step back a bit and look at the problem from a high level perspective, then I think it should be acceptable to require callers of patch/replace operations to provide _rev in the data, as that's the way how CouchDB/Cloudant allows clients to detect and resolve conflicts

Our current implementation:
Now whenever `rev` is needed, eg: in function `Cloudant.prototype.delete()`, `Cloudant.prototype.replaceById()`, we call `getCurrentRevision()`.
But considering the nature of Cloudant db, we should ask the caller to provide `rev`, and return error if it's missing.

Affected functions:
  - updateOrCreate
    - PATCH Models/
    - Provide rev when updating an existed instance with `data: {rev: rev_of_new_data} `
  - replaceOrCreate
    - PUT Models/
    - Provide rev when replacing an existed instance with `data: {rev: rev_of_new_data} `
  - save
    - need to investigate
  - replaceById
    - Provide rev when replacing an existed instance with `data: {rev: rev_of_new_data} `
  - delete
    - DELETE Models/
    - We disabled this endpoint and DELETE Models/{id} doesn't accept data...any idea? Probably keep our current design.

bajtos:
> But considering the nature of Cloudant db, we should ask the caller to provide rev, and return error if it's missing.

A big +1 from me.

I believe the field is called `_rev` in Cloudant, isn't it? https://docs.cloudant.com/mvcc.html#revisions

I am proposing to define a `_rev` property on a model at the time when the model is attached to a Cloudant datasource. That way the `_rev` property will be propagated to Swagger too.

The catch is how to tell clients that `_rev` should not be included in `create` requests? This is similar to the problem of excluding `id` property when `forceId` is `true`, see https://github.com/strongloop/loopback/issues/2924

jannyHou:
@bajtos ah sorry I forgot to direct you to https://github.com/strongloop/loopback-connector-cloudant/pull/54#issuecomment-260415145, I made some comment about how to let callers access `_rev` in that link. 

> The catch is how to tell clients that _rev should not be included in create requests

True, my suggestion is based on using a option flag to tell function `fromDB()` returns `_rev` with data or not.

kjdelisle:
@ssh24 What's left for this task:
[...]

PR #123, opened 2017-05-05
allow handling of ._rev on models

kjdelisle:
### Description

connected to https://github.com/strongloop/loopback-connector-cloudant/issues/55
By letting the application developer decide how to handle document conflict errors, we sidestep the potential problems of enforcing a document conflict resolution strategy.


kjdelisle:
@slnode test please

b-admike:
@slnode test please

b-admike:
@slnode test please

kjdelisle:
@slnode test please

kjdelisle:
@slnode test please

kjdelisle:
@slnode test please

kjdelisle:
@slnode test please

kjdelisle:
@slnode test please

ssh24:
Please note travis will fail. It is because `replaceById` returns back the old `_rev` property and our test case catches it. The fix is on juggler https://github.com/strongloop/loopback-datasource-juggler/pull/1378

README.md
```diff
@@ -164,9 +163,117 @@ modelSelector | JSON | Use the Cloudant Query selector syntax to associate model
 
 ### _rev Property
 
-In a document, property `_rev` is the latest doc revision and must be provided when modifying the doc. Currently when updating/deleting a doc, the connector first retrieves that doc, then gets its `_rev`, then calls a modify api with `_rev` included. It makes three calls to the database and this would easily result in race condition.
-  
-Considering the nature of Cloudant, we plan to give user the flexibility to define `_rev` property in model and provide `_rev` by themselves instead of the connector. In such way user can get the revision and control conflict before using that revision to modify the document.
+In a document, property `_rev` is the latest doc revision and must be provided when modifying the doc.
```
kjdelisle: Why is there a ton of whitespace here?

README.md
```diff
@@ -164,9 +163,117 @@ modelSelector | JSON | Use the Cloudant Query selector syntax to associate model
...
+
+Our connector allows the user to retrieve back the `_rev` property upon all CRUD operations, however does not add it to the model definition.
+
```
kjdelisle: remove this line

README.md
ssh24: There are no white spaces ...
Check the https://github.com/strongloop/loopback-connector-cloudant/tree/allow-underscore-rev#_rev-property

Just the way github shows the diff.

README.md
```diff
@@ -164,9 +163,117 @@ modelSelector | JSON | Use the Cloudant Query selector syntax to associate model
...
+      id:'2',
+      _rev:'2-abcedf',
+      name:'Bar4'
+    }, function(err, result) {
+      if (err) throw err;
+      console.log('Update an existing instance: ' + JSON.stringify(result));
+    });
+  ```
+
+**Note:** When calling either `replaceOrCreate` or `updateOrCreate` function, **_do not specify the `_rev` value**_, unless the user is confident it should do a `replace`/`update` instead of a `create`. This is because Cloudant does not allow custom `_rev` value on a `create` operation.
```
kjdelisle: I think in this particular case, we should be handling this in code.
If we're attempting to update the document, and it's returning a 404 (with a provided `_.rev` value) then we should be stripping that `_.rev` off and attempting to create it.

lib/cloudant.js
```diff
@@ -742,23 +736,41 @@ Cloudant.prototype.updateOrCreate = function(model, data, cb) {
...
+  };
+
   if (id) {
-    self.updateAttributes(model, id, data, {}, function(err, docs) {
-      if (err && err.statusCode && err.statusCode === 404) {
-        self.create(model, data, {}, function(err) {
+    self.exists(model, id, {}, function(err, count) {
+      if (err) return cb(err);
+      else if (count > 0) {
+        self.getCurrentRevision(model, id, function(err, rev) {
```
kjdelisle: This is completely wrong, and **not** what I had in my PR originally! This will overrwrite the provided `_.rev` from the user! :-1:

kjdelisle: As far as I'm concerned, the getCurrentRevision function is completely superfluous and should be removed from the API entirely. It's only purpose is to confuse us, and our users.

If the user wants an updated version of the document with the latest `_.rev`, then they should be running `find` or `findById`.

lib/cloudant.js
```diff
@@ -822,28 +834,38 @@ Cloudant.prototype.replaceOrCreate = function replaceOrCreate(
...
   if (id) {
-    self.getCurrentRevision(model, id, function(err, rev) {
+    self.exists(model, id, {}, function(err, count) {
       if (err) return cb(err);
-      if (rev) data._rev = rev;
-      self._insert(model, data, function(err) {
-        if (err) return cb(err);
-        mo.db.get(id, function(err, doc) {
+      else if (count > 0) {
+        self.getCurrentRevision(model, id, function(err, rev) {
```
kjdelisle: :-1:

Again, this function defeats the purpose of my PR. It only works because it's hiding how Cloudant should work.

lib/cloudant.js
```diff
@@ -862,19 +884,25 @@ Cloudant.prototype.replaceById = function replaceById(
...
-        if (err) return cb(err);
-        cb(null, self.fromDB(model, mo, doc));
-      });
+      cb(null, self.fromDB(model, mo, doc));
     });
-  });
+  };
+
+  if (!data._rev) {
+    self.getCurrentRevision(model, id, function(err, rev) {
```
kjdelisle: See above comments, rework this to remove this function call.

kjdelisle:
Apparently, I'm not allowed to "request changes" on my review for some reason.

README.md
kjdelisle: I should point out that this is probably the only case where we want to "do a smart thing" for the user.

README.md
rmg: This is what happens when you don't limit your files to 80 columns ;-)

ssh24: @rmg How do we enforce those rules on a Markdown file?

README.md
kjdelisle: @ssh24 ^^^

lib/cloudant.js
```diff
@@ -742,23 +736,32 @@ Cloudant.prototype.updateOrCreate = function(model, data, cb) {
...
+
   if (id) {
     self.updateAttributes(model, id, data, {}, function(err, docs) {
-      if (err && err.statusCode && err.statusCode === 404) {
-        self.create(model, data, {}, function(err) {
+      if (err && err.statusCode !== 404) return cb(err);
+      else if (err && err.statusCode && err.statusCode === 404) {
+        self.create(model, data, {}, createHandler);
+      } else {
+        mo.db.get(id, function(err, doc) {
```
kjdelisle: Does `updateAttributes` not return the resulting document after it's been changed? If it doesn't, then let's fix that instead.

Unless of course there's a Juggler-level API implication surrounding `updateAttributes`, where we don't give back the whole record again. :/

kjdelisle:
`%s/_.rev/._rev/g`
special thanks to @rmg
XD

README.md
ssh24: @kjdelisle Fixed this. Now the user can call `replaceOrCreate`, `updateOrCreate`, `replaceById`, `update` and send out a _rev value. We strip it out before making the final _insert call.

ssh24:
> I wrote the test-util hasError method to assert against having both err and result come back at the same time. Apparently, there's at least one juggler function where a 409 Conflict error will still return the result object (which is just the document you originally submitted in an update attempt). In those cases, you should make sure you avoid using the helper checks, and just assert that the error exists and has the right code and such.

@kjdelisle Are you talking about `replaceById` function where it returns the old `_rev`? I do not understand your concern then. The way Juggler does it is wrong and the test case fails in the correct way. The fix on Juggler is [here](https://github.com/strongloop/loopback-datasource-juggler/pull/1378). We need to land that PR first before proceeding with this one.

README.md
```diff
@@ -164,9 +163,115 @@ modelSelector | JSON | Use the Cloudant Query selector syntax to associate model
...
+    name: 'Foo',
+  }, {
+    name: 'Bar',
+  }], function(err, result) {
+    if (err) throw err;
+    console.log('Created instance: ' + JSON.stringify(result));
+  });
+```
+
+**Note:** Cloudant does not allow customized `_rev` value, hence creating an instance with a `_rev` value will give an error. The onus is on the user if they fail to comply to this rule.
```
b-admike: You can warn them, but how does it result in an error if we strip the `_rev` value even if they provide it? i.e. 
```
 // strip the _rev property out since users should not update the _rev value either ways
+  delete data._rev;
```

ssh24:
Interestingly, stripping out the `_rev` value does for update/replace does not work. Juggler stores the `_rev` value out of the context of `_data` and then returns it back to the connector. Now this breaks because even if the user did not do a `replaceOrCreate` with a `_rev` value, it will try to strip out the default `_rev` and break!

As a result, reverting the last commit.

README.md
ssh24: @kjdelisle I cannot think of a way to get past this situation. If the user does not specify the _rev value, Juggler automatically gets it. So if we are stripping out in the connector, we possibly strip out the **actual** value even when we didn't intend to do so.

test/update.test.js
```diff
@@ -0,0 +1,158 @@
...
+      Product.updateOrCreate(updatedBread, function(err, result) {
+        should.not.exist(err);
+        should.exist(result._rev);
+        testUtil.checkModel(updatedBread, result);
+        done();
+      });
+    });
+  });
+
+  it('throws on update when model exists and _rev is different ',
```
b-admike: How is `throws on update when model exists and _rev is different` different than `throws on replace when model exists and _rev is different`? They both use the `create` method and same set of logic from what I can tell.

ssh24: I was confused by that as well. @kjdelisle Added that set of tests and they are duplicate across `test/update.test.js`, `test/replace.test.js` and `test/create.test.js`

README.md
b-admike: ^ I was wondering where all the stripping code went and why.

ssh24: @b-admike I reverted that commit where we were stripping out the `_rev`

kjdelisle: @ssh24 That doesn't make sense. If you attempt whatever the operation is and it fails (you attempt to update, and Cloudant returns an error because there's nothing to update), then you would just remove the `._rev` at that point and attempt creation.

test/update.test.js
kjdelisle: Erm...they didn't used to do that. The only time create was used in my original PR was as the first operation in the chain. @ssh24 Are you sure you didn't search-replace that command name?

ssh24: Just realized that. I will cherry pick those tests back from your commit.

lib/cloudant.js
ssh24: @kjdelisle `replaceById` apparently does not store the `_rev` value like other `replaceOrCreate` or `updateOrCreate` method. We might have to just keep the `getCurrentRevision` or fix the juggler dao.

kjdelisle:
@ssh24 I figured out why I can't review; I'm the author of the PR. XD

I left some additional comments.

README.md
ssh24: Fixed.

README.md
rmg: @ssh24 I'm not sure I understand the question. All PRs are reviewed, aren't they? Or do you mean something like a linter?

ssh24: @rmg Yes, a linter for README or any markdown files.

lib/cloudant.js
```diff
@@ -712,7 +706,10 @@ Cloudant.prototype.updateAttributes = function(model, id, data, options, cb) {
     _.mergeWith(doc, data, function(dest, src) { return src; });
     self.create(model, doc, options, function(err) {
       if (err) return cb(err);
-      return cb(err, data);
+      mo.db.get(id, function(err, doc) {
```
jannyHou: any reason why function `_insert` only returns `id` but not the full `result`? I think the `result` returned from database is the best data to show the insert result, `mo.db.get` would have delay, which lead to 409 conflict if that doc get updated by other one.

ssh24: It was always returning `id` by default and returned the whole inserted instance to the user (I think dao just appends back the `id`). Now it returns `id` and `_rev` and the user still gets back the whole instance.

rmg: If this is calling `db.insert()` behind the scenes then there is no reason to perform an additional `db.get()`. Simply add the returned `_rev` to the local copy of the document and use that. CouchDB/Cloudant does not support partial updates, only whole-document updates, so any operation to update a document involves constructing the full document locally and then pushing it to the DB.

ssh24: That's the issue me and @jannyHou were just talking about. `_insert` returns back only the `_rev` and `id` property and not the whole instance.

Why it does that? Because `.create()` in dao stores the to be inserted data and once the call from `_insert` is returned, it just appends the `id` and `_rev` value back to the stored data and sends it back to the user.

This makes using `_insert` function not so useful because one might think they are getting back the whole instance that got inserted but instead they get back the `id` and `_rev`.

jannyHou: @rmg I talked with Sakib, connector function(in this case, `create`) only callback the `id` returned from database, then in juggler, we merge `id` and the data user provide in payload, then callback it to user. I bring up the question in slack channel, I am confused with the reason behind this design...

jannyHou: @ssh24 regardless what juggler does with the data, we can still callback the full `result` in `_insert`, it's a connector private function and won't be used for juggler.

If `Cloudant.prototype.create()` expect only `id` returned, then let itself do the result filter, and here in `updateAttributes()`, we can call `self._insert` instead of `self.create`, unless there are required check in `create()`

rmg: This doesn't explain why `mo.db.get()` has to be called here when it will give the exact same document as is already in the `doc` variable - except for `doc._rev` needing to be updated, which should have been captured when `insert()` was called.

ssh24: Ah, I see. Agree with that. Fixed it.

README.md
kjdelisle: @ssh24 This extra line is still here...

lib/cloudant.js
```diff
@@ -742,23 +737,33 @@ Cloudant.prototype.updateOrCreate = function(model, data, cb) {
...
+
   if (id) {
     self.updateAttributes(model, id, data, {}, function(err, docs) {
-      if (err && err.statusCode && err.statusCode === 404) {
-        self.create(model, data, {}, function(err) {
-          if (err) return cb(err);
-          return cb(err, data, {isNewInstance: true});
-        });
+      if (err && err.statusCode !== 404) return cb(err);
+      else if (err && err.statusCode && err.statusCode === 404) {
```
kjdelisle: If `err.statusCode` didn't exist, then the if statement would've been true and it would've returned.
Checking `err.statusCode` should be fine at this point.

lib/cloudant.js
```diff
@@ -862,19 +871,25 @@ Cloudant.prototype.replaceById = function replaceById(
   var self = this;
   var mo = self.selectModel(model);
   var idName = self.idName(model);
-  data[idName] = id;
+  data[idName] = id.toString();
```
kjdelisle: :+1:
I missed that bit myself.

README.md
ssh24: What extra line? If you see the line numbers on the side, we have `Our connector allows the user to retrieve back the `_rev` property upon all CRUD operations, however does not add it to the model definition.` on L168 and then space out  L169 empty and write the next sentence on L170.

lib/cloudant.js
kjdelisle: Then fix the DAO we must.
On the face of it, `replaceById` is one of those functions where the user might say "Ehh, I don't care about what's there. If it's there, get rid of it".
The trouble is that if someone were to update that document before the user does, how do we, the framework devs of LoopBack, know whether or not a user doesn't want to halt at that point and rethink their choice?

For the same reason as `updateOrCreate` and `replaceOrCreate`, I don't think that we should have `replaceById` altering data with mismatched `_rev` values.

lib/cloudant.js
ssh24: Agree. Fixed.

README.md
kjdelisle: Ah, it's another symptom of the line wrap.

kjdelisle:
LGTM

README.md
```diff
@@ -164,9 +163,117 @@ modelSelector | JSON | Use the Cloudant Query selector syntax to associate model
...
-  
-Considering the nature of Cloudant, we plan to give user the flexibility to define `_rev` property in model and provide `_rev` by themselves instead of the connector. In such way user can get the revision and control conflict before using that revision to modify the document.
+In a document, property `_rev` is the latest doc revision and must be provided when modifying the doc.
+
+Our connector allows the user to retrieve back the `_rev` property upon all CRUD operations, however does not add it to the model definition.
+
+If you would like to have a `_rev` property on your model, as an end user, the onus is on you to add the property in the model definition.
+
+**Note:** All CRUD operations require `_rev` (except create) and __it is upto the user to specify them__. The connector does not handle such cases due to possibilities of race condition when two users try to update the same document.
```
0candy: `upto` should e two words

kjdelisle:
@ssh24 Thanks for all your effort, and for your patience with my endless nitpicking. ^^

README.md
ssh24: @0candy Fixed.

PR #128, opened 2017-05-09
Fix init config

ssh24:
Fix init config to get latest config from loopbackrc


kjdelisle:
@slnode test please

kjdelisle:
We need to make sure that the rc loader doesn't run if we're on Travis CI (since that will never work).

ssh24:
@slnode test please

ssh24:
This PR is not needed. CIS-jenkins config has been updated to use the same ones set in loopbackrc.

PR #129, opened 2017-05-09
1.x fix/upsert

jannyHou:
related to https://github.com/strongloop/loopback-connector-cloudant/pull/111

### DIFF
--- a/test/cloudant.test.js
+++ b/test/cloudant.test.js
@@ -115,10 +115,7 @@ describe('cloudant connector', function() {
 
   describe('updateOrCreate', function() {
     afterEach(function(done) {
-      Product.destroyById(1, function(err) {
-        if (err) return done(err);
-        done();
-      });
+      Product.destroyById(1, done);
     });
 
     it('can call upsert with an id and have it create a document without error',
@@ -131,11 +128,12 @@ describe('cloudant connector', function() {
           if (err) return done(err);
           res.id.should.equal('1');
           res.name.should.equal('bread');
-          // scenario: upsert does a update
+          // scenario: upsert does an update
           Product.upsert({
             id: '1',
             name: 'newBread',
           }, function(err, res) {
+            if (err) return done(err);
             res.id.should.equal('1');
             res.name.should.equal('newBread');
             done();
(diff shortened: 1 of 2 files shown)


jannyHou:
@slnode test please

test/cloudant.test.js
```diff
@@ -112,6 +112,37 @@ describe('cloudant connector', function() {
         };
       });
   });
+
+  describe('updateOrCreate', function() {
+    afterEach(function(done) {
+      Product.destroyById(1, function(err) {
+        if (err) return done(err);
+        done();
```
Amir-61: Instead of:
```
if (err) return done(err);
done();
```

Just the following line is enough:
```
done(err)
```

test/cloudant.test.js
```diff
@@ -112,6 +112,37 @@ describe('cloudant connector', function() {
...
+      function(done) {
+        // scenario: upsert does a create
+        Product.upsert({
+          id: '1',
+          name: 'bread',
+        }, function(err, res) {
+          if (err) return done(err);
+          res.id.should.equal('1');
+          res.name.should.equal('bread');
+          // scenario: upsert does a update
```
Amir-61: **an** update

test/cloudant.test.js
```diff
@@ -112,6 +112,37 @@ describe('cloudant connector', function() {
...
+          Product.upsert({
+            id: '1',
+            name: 'newBread',
+          }, function(err, res) {
```
Amir-61: Please check `err` here as well.

test/cloudant.test.js
```diff
@@ -112,6 +112,37 @@ describe('cloudant connector', function() {
...
+
+  describe('updateOrCreate', function() {
+    afterEach(function(done) {
+      Product.destroyById(1, function(err) {
+        if (err) return done(err);
+        done();
+      });
+    });
+
+    it('can call upsert with an id and have it create a document without error',
```
Amir-61: Probably the following is what you are achieving here:

```
can call upsert with an a primary key and create/update a document without error
```

Also it is always a good idea if you try primary keys which are not `id` as well.

PR #131, opened 2017-05-11
cloudant.test: cleanup after test runs

kjdelisle:
### Description
The current tests aren't properly cleaning up after themselves, this PR should solve the problem.

#### Related issues
Connect to https://github.com/strongloop-internal/scrum-apex/issues/204

### DIFF
--- a/test/cloudant.test.js
+++ b/test/cloudant.test.js
@@ -12,6 +12,8 @@ var url = require('url');
 var should = require('should');
 var db, Product, CustomerSimple, SimpleEmployee;
 
+var QUERY_MAX = 1000;
+
 describe('cloudant connector', function() {
   before(function(done) {
     db = getDataSource();
@@ -71,10 +73,7 @@ describe('cloudant connector', function() {
 
   describe('replaceOrCreate', function() {
     after(function cleanUpData(done) {
-      Product.destroyAll(function removeModelInstances(err) {
-        if (err) return done(err);
-        done();
-      });
+      Product.destroyAll(done);
     });
     it('should replace a model instance if the passing key already exists',
       function(done) {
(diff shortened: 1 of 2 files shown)


jannyHou:
@kjdelisle It's good to put `limit` into `options`, probably you need to change code in [`Cloudant.prototype.all`#L271](https://github.com/strongloop/loopback-connector-cloudant/blob/master/lib/cloudant.js#L271) to add `query.limit` from `options` as well.

I could not guarantee increase `query.limit` will return you all existing instances, you may want to hardcode its value to 1000 and see if CI passes. If that's the fix, then continue your implementation.

kjdelisle:
@jannyHou I wouldn't expect the 1000 to be a magic fix, especially if we were testing with more than 1000 rows. I selected that limit as a default for the tests where keeping a mysterious count of how many items went into the database would be sketchy or brittle.

Many of our describe sections don't handle their tests by pre-creating a fixed length array of items to work with, so "knowing" the count would mean counting up all the non-overlapping records that may have been made by the individual test calls themselves. This would mean that any time anyone wanted to add a new test, they would have to change that magic number to include whatever extra records they were adding.

There is one section where I base the limit off the pre-seeded array size. If that was the norm and not the exception, then I would do that instead.

package.json
```diff
@@ -16,6 +16,7 @@
   "main": "index.js",
   "scripts": {
     "lint": "eslint .",
+    "pretest": "sh setup.sh",
```
kjdelisle: One point of note: The .travis.yml already does this. If you're doing it here, you need to remove it from [here](https://github.com/strongloop/loopback-connector-cloudant/blob/master/.travis.yml)

jannyHou:
was about to give LGTM :p 
LGTM

Issue #67, opened 2016-12-12
Support to query view


xi4oc4i:
### Bug or feature request

- [ ] Bug
- [x ] Feature request

### Description of feature (or steps to reproduce if bug)

Couch connector supports to query the view, will cloudant connector support this feature?

https://www.npmjs.com/package/loopback-connector-couch

dhmlau:
@raymondfeng, any thought on supporting query on view (or search indexes) for Cloudant connector?

jannyHou:
query view would be a cloudant specific function, it will be good if we can define it in cloudant connector, like:
```
// in `lib/cloudant.js`
Cloudant.prototype.queryView() {
 // code
}
```
then in juggler `lib/datasource.js`, we can mix certain connector specific functions to Model, like how we mix DAO functions to model in https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/datasource.js#L585

This would also be useful across all connectors.
@raymondfeng Any thought?

dhmlau:
@jannyHou @raymondfeng , I think it's a good idea to have a consistent story in supporting views across connectors because they are not specific to Cloudant.  

Similar question in mssql: https://github.com/strongloop/loopback-connector-mssql/issues/26

jannyHou:
- [db.view(designname, viewname, [params], [callback])](https://github.com/apache/couchdb-nano#dbviewdesignname-viewname-params-callback)
  - This is the main function exported by nano to query view.
  - If our goal is to support it, we can define a method `Cloudant.prototype.view(viewname, [params], cb)`, `designname` can be read from loopback model name, `viewname` and `params` are inputs.

jannyHou:
Turns our the `view` could not be a model level api with current design, explained in https://github.com/strongloop/loopback-connector-cloudant/issues/15#issuecomment-300955117


Other plans:

- strategy1: Support those design functions like view as datasource level functions, so a view don't need to be created in a lb-model's design doc.
- srategy2: Use two design docs(one for index one for view) to represent a loopback model.

b-admike:
Steps taken to verify:

- Create a DDoc on cloudant db called `TestViewDDoc` with a view `hadtraining`
```json
{
  "_id": "_design/TestViewDDoc",
  "_rev": "13-2738c014de2d5b6cfc02522d1bdc45fc",
  "views": {
    "hadtraining": {
[...]

PR #133, opened 2017-05-12
viewDocs

jannyHou:
### Description
connect to strongloop/loopback-connector-cloudant#67

- [x] Add a datasource level function called `viewDocs` which does view query
- [x] Add test cases

Please note: 
- View cannot be created in a loopbackModel's design doc, it would be some ddoc that already exists in user's database.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,11 +37,11 @@ The `loopback-connector-cloudant` module is the Cloudant connector for the LoopB
     - [Example Code](#example-code)
 - [Discovery](#discovery)
 - [Query](#query)
+- [View](#view)
 - [Testing](#testing)
         - [Docker](#docker)
 - [More Info](#more-info)
 - [Feature backlog](#feature-backlog)
-    - [View](#view)
     - [Index](#index)
 
 <!-- /TOC -->
@@ -531,6 +531,37 @@ Not implemented yet, track it in story https://github.com/strongloop/loopback-co
 -  [LoopBack query](http://loopback.io/doc/en/lb3/Querying-data.html) support for: fields, limit, order, skip and where filters.
 - Please check [Advanced Queries](https://github.com/strongloop/loopback-connector-cloudant/blob/master/doc/advanced-queries.md) for details about regex filter, nested filter and order.
 
+# View
+
+Given a design doc name and the view name in it, user can use a connector level function `viewDocs` to query the view.
+
+Since `viewDocs` is a specific api for Cloudant connector only, it is not attached to the dataSource Object defined in loopback-datasource-juggler, which means the correct way to call it is `ds.connector.viewDocs`:
+
+*/server/script.js*
+```javascript
+module.exports = function(server) {
+  // Get Cloudant dataSource as `ds`
+  // 'cloudantDB' is the name of Cloudant datasource created in 
+  // 'server/datasources.json' file
+  var ds = server.datasources.cloudantDB;
+
+
+  ds.on('connected', function() {
+    // 1. Please note `ds.connector.viewDocs()` is the correct way to call it,
+    // NOT `ds.viewDocs()`
+    // 2. This api matches the Cloudant endpoint:
+    // GET /db/_design/<design-doc>/_view/<view-name>
+    ds.connector.viewDocs('design_doc', 'view_name', function(err, results) {
+      // `results` would be the data returned by quering that view
+    });
+
+    // Alternatively user can also specify the filter for view query
+    ds.connector.viewDocs('design_doc', 'view_name', {key: 'filter'}, 
+      function(err, results) {});
+  });
+};
+```
+
 # Testing
 
 - Cloudant local(docker image)
@@ -584,10 +615,6 @@ see the [docs section](https://github.com/strongloop/loopback-connector-cloudant
 * Index-only model properties marked with index=true
 * Configurable "view based" or JSON indexes. [More Info>>](https://cloudant.com/blog/mango-json-vs-text-indexes)
 * [Support uuid](https://github.com/strongloop/loopback-connector-cloudant/issues/104)
-
-## View
-
-Not implemented yet, track it in https://github.com/strongloop/loopback-connector-cloudant/issues/67
 
 ## Index
 
--- a/lib/view.js
+++ b/lib/view.js
@@ -1,0 +1,55 @@
+// Copyright IBM Corp. 2015,2016. All Rights Reserved.
+// Node module: loopback-connector-cloudant
+// This file is licensed under the Artistic License 2.0.
+// License text available at https://opensource.org/licenses/Artistic-2.0
+
+'use strict';
+
+var URL = require('url');
+var assert = require('assert');
+var util = require('util');
+var _ = require('lodash');
+
+module.exports = mixinView;
+
+function mixinView(Cloudant) {
+  var debug = require('debug')('loopback:connector:cloudant:view');
+
+  /**
+   * Get data returned by querying a view
+   *
+   * @param {String} ddocName The design doc name
+   * @param {String} viewName The view name
+   * @param {Object} options The cloudant view filter
+   * @param {Function} [cb] The callback function
+   */
+  Cloudant.prototype.viewDocs = function(ddocName, viewName, options, cb) {
+    // ds.viewDocs(ddocName, viewName, cb);
+    if (typeof options === 'function' && !cb) {
+      cb = options;
+      options = {};
+    }
+    debug('Cloudant.prototype.view ddocName %s viewName %s options %s',
+      ddocName, viewName, options);
+
+    var dbName = this.settings.database || this.settings.db ||
+      getDbFromUrl(this.settings.url);
+    var db = this.cloudant.use(dbName);
+    assert(db && typeof db.view === 'function',
+      'model is not attached to a datasource supporting function view');
+
+    db.view(ddocName, viewName, options, cb);
+  };
+};
+
+/**
+ * Parse url and return the database name if provided
+ * @param {String} url The cloudant connection url
+ * @return {String} The database name
+ */
+function getDbFromUrl(url) {
+  var parsedUrl = URL.parse(url);
+  if (parsedUrl.path && parsedUrl.path !== '/')
+    return parsedUrl.path.split('/')[1];
+  else return '';
+}
--- a/lib/view.js
+++ b/lib/view.js
@@ -1,4 +1,4 @@
-// Copyright IBM Corp. 2015,2016. All Rights Reserved.
+// Copyright IBM Corp. 2017. All Rights Reserved.
 // Node module: loopback-connector-cloudant
 // This file is licensed under the Artistic License 2.0.
 // License text available at https://opensource.org/licenses/Artistic-2.0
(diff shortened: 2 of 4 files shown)


jannyHou:
CI passed(1st commit), now it's running due to a jsdoc change.

lib/view.js
```diff
@@ -0,0 +1,55 @@
...
+function mixinView(Cloudant) {
+  var debug = require('debug')('loopback:connector:cloudant:view');
+
+  /**
+   * Get data returned by querying a view
+   *
+   * @param {String} ddocName The design doc name
+   * @param {String} viewName The view name
+   * @param {Object} options The cloudant view filter
+   * @param {Function} [cb] The callback function
```
b-admike: Do we need to follow the `@callback` format here for JSDocs?

lib/view.js
```diff
@@ -0,0 +1,55 @@
...
+   */
+  Cloudant.prototype.viewDocs = function(ddocName, viewName, options, cb) {
+    // ds.viewDocs(ddocName, viewName, cb);
```
b-admike: intentional comment?

jannyHou: yep :) the following two lines are written to make inputs compatible with the scenario in that comment.

lib/view.js
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * Parse url and return the database name if provided
+ * @param {String} url The cloudant connection url
+ * @return {String} The database name
+ */
+function getDbFromUrl(url) {
+  var parsedUrl = URL.parse(url);
+  if (parsedUrl.path && parsedUrl.path !== '/')
+    return parsedUrl.path.split('/')[1];
+  else return '';
```
ssh24: Don't need an `else` statement here. Just return.

lib/view.js
```diff
@@ -0,0 +1,56 @@
+// Copyright IBM Corp. 2015,2016. All Rights Reserved.
```
rmg: Should just be `2017`, not `2015,2016`.

lib/view.js
```diff
@@ -0,0 +1,56 @@
...
+
+module.exports = mixinView;
+
+function mixinView(Cloudant) {
+  var debug = require('debug')('loopback:connector:cloudant:view');
+
+  /**
+   * Get data returned by querying a view
+   *
+   * @param {String} ddocName The design doc name
```
rmg: What form is the `ddocName` expected to be in?
 - `/{db}/_design/{name}`
 - `_design/{name}` (db is implied)
 - `{name}` (db and `_design/` is implied)

test/cloudant.view.test.js
```diff
@@ -0,0 +1,97 @@
...
+      };
+      function insertViewDdoc(cb) {
+        // The design doc must be stored with double quote,
+        // and map function cannot join by '+'
+        /* eslint-disable */
+        var ddoc = {
+          "_id": "_design/getModel",
+          "views": {
+            "returnModelInstances": {
+              "map": "function(doc) { if(doc.loopback__model__name) { emit(doc.loopback__model__name, doc); }}"
```
rmg: You might find [Function.prototype.toString()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/toString) helpful here ;-)

test/cloudant.view.test.js
```diff
@@ -0,0 +1,97 @@
...
+
+    it('returns result by query a view', function(done) {
+      db.connector.viewDocs('getModel', 'returnModelInstances',
+        function(err, results) {
+          results.total_rows.should.equal(5);
+          results.rows.forEach(isLBModelInstance);
+          done(err);
+
+          function isLBModelInstance(elem) {
+            elem.value.hasOwnProperty('loopback__model__name').
```
rmg: If the intended use-case for this feature is existing databases with existing design docs, I would advise against using property names like `loopback__model__name` for a couple reasons:
 1. it might make code readers/reviewers think this is a field added by LoopBack
 2. it might make current/future maintainers think this is an official property name

Adding a property to every document in an existing CouchDB database is a very expensive operation. So expensive it might even be considered impossible by some operators. The only reasonable approach would be to have the user provide the map function and/or property name _already_ being used for this purpose. If that isn't considered upfront, it could lead to assumptions and design decisions that will need to be undone later and may require future breaking changes.

rmg: To be more specific, I would use something like `test_model_name` or similar for the property name so that it is more clear that it is _not_ an official/magic string.

lib/view.js
jannyHou: It is the third one, `{name} (db and _design/ is implied)`, and this is also what the driver function  `mo.db.viewDocs()` expects.

lib/view.js
kjdelisle: Does the copyright tool make those changes for us, now? We can always run a separate PR to update it.

lib/view.js
kjdelisle: +1
The JSDoc should explain which format is expected.

lib/view.js
```diff
@@ -0,0 +1,56 @@
...
+   * @param {String} viewName The view name
+   * @param {Object} options The cloudant view filter
+   * @callback
+   * @param {Function} cb
+   */
+  Cloudant.prototype.viewDocs = function(ddocName, viewName, options, cb) {
+    // ds.viewDocs(ddocName, viewName, cb);
```
kjdelisle: Remove commented code.

lib/view.js
rmg: Fixing the JSDoc comment to answer my question was what I intended.

lib/view.js
rmg: It does, yes. The tool could also just be run on this branch and fixed as part of this PR ;-)

lib/view.js
kjdelisle:
```
/**
 * Gets data at `/{db}/_design/{ddocName}/views/{viewName}`
 * @param {string} ddocName The name of the design document.
 * @param {string} viewName The name of the view function attached to the design document.
 * //etc
 */
```

test/cloudant.view.test.js
kjdelisle: I would actually prefer it, to be honest. We should make that a default expectation so that ESLint can protect us.

test/cloudant.view.test.js
jannyHou: When I wrote the test I didn't think into that deep, I was just thinking of a way to simply insert data (LBModel.create()), it doesn't imply "Adding a property to every document in an existing CouchDB database".
The view could be used for non-loopback data(already exist in database), and also could be used to improve LB model's performance.

Your two concerns are reasonable to me, I will change the field to some name that not misleading.

test/cloudant.view.test.js
```diff
@@ -0,0 +1,97 @@
...
+// License text available at https://opensource.org/licenses/Artistic-2.0
+
+'use strict';
+
+require('./init.js');
+var Cloudant = require('../lib/cloudant');
+var _ = require('lodash');
+var url = require('url');
+var should = require('should');
+var async = require('async');
```
kjdelisle: Sort library imports, please. :)

lib/view.js
jannyHou: @kjdelisle ah, I rephrased the comments, it was confusing.

jannyHou:
@kjdelisle @rmg Changes applied, PTAL again. Thanks.

lib/view.js
```diff
@@ -0,0 +1,73 @@
...
+    if (typeof options === 'function' && !cb) {
+      cb = options;
+      options = {};
+    }
+    debug('Cloudant.prototype.view ddocName %s viewName %s options %s',
+      ddocName, viewName, options);
+
+    var self = this;
+    var db = this.cloudant.use(self.getDbName(self));
+    assert(db && typeof db.view === 'function',
```
rmg: This isn't checking the design doc or the datasource. `this.cloudant.use` is just returning a scoped object so that methods on it use URLs based on the db name. If `db.view` is ever not a function then it would mean that someone has changed the implementation of this connector to no longer use `cloudant`/`cloudant-nano`/`nano` as the client.

Is that the purpose of this assertion?

lib/view.js
```diff
@@ -0,0 +1,73 @@
...
+   * Gets data at `/{db}/_design/{ddocName}/views/{viewName}`
+   *
+   * Example:
+   * User has a view called `getModel` in design document /{db}/_design/model,
+   * to query the view, user can call function:
+   * ```
+   * ds.viewDocs(model, getModel, {key: 'purchase'}, cb);
+   * ```
+   *
+   * @param {String} ddocName The design doc name
```
jannyHou: @rmg I am confused why a user would think of using part of the url as a design doc name? By reading couchdb/cloudant document, or those nodejs drivers' README.md, I don't see anywhere implies that the design document name would contain the those prefix....

lib/view.js
jannyHou: I thought if the dynamically loaded database doesn't exist, then `this.cloudant.use('dbName')` would return undefined, but I just checked it throws error saying db doesn't exist. 
Ok I will remove that assert.

test/cloudant.view.test.js
```diff
@@ -0,0 +1,104 @@
...
+      });
+    });
+  });
+});
+
+function generateSamples() {
+  var samples = [
+    {
+      model: 'purchase',
+      'customer_id': '1',
```
jannyHou: nope :p  changed it to `Number`.

jannyHou:
Thanks @rmg I applied the other two comments, and explained my confusion of the "design doc name" in https://github.com/strongloop/loopback-connector-cloudant/pull/133#discussion_r118972236, I am ok to add the prefix explanation in the description if you insist, just curious why it's that important, even after having an example usage :)

lib/view.js
rmg: As a long time CouchDB user, _I am_ confused by whether I am expected to provide the whole name or not. Since the actual Couch/Cloudant document name is actually `_design/<name>` (with the `/` requiring URL encoding as `%2F`), it makes sense to _me_ to pass in the full name - unless I have written my own wrapper which adds it.

It is a case where knowing more means things are harder rather than easier.

lib/view.js
rmg: That sort of dynamic check would require network I/O, which would require `.use()` to be async and take a callback ;-)

rmg:
Paraphrasing my reply from the line-comment since I think it is important enough to not hide:
The actual name of the document, which is the "foo" design document is `_design/foo`. This is a normal Couch/Cloudant/JSON document, but the name includes the prefix `_design/` which CouchDB/Cloudant uses as a filter for documents which contain views, lists, etc. Since `/` is part of the actual document's name, it is _sometimes_ necessary to URL encode it so that it doesn't get treated as a path separator.

Yes, this is a little weird, and yes, it is sort of like having a file on disk with `/` in the name.

jannyHou:
@rmg Thank you for elaborating the reason, by
> the actual Couch/Cloudant document name is actually _design/<name> (with the / requiring URL encoding as %2F

you are right 👍  
And I double checked, `GET /db/_design/<design-doc>/_view/<view-name>` is the correct endpoint.
Changes applied. PTAL again. Thanks!

test/cloudant.view.test.js
rmg: Actually I meant the key it self.

test/cloudant.view.test.js
```diff
@@ -0,0 +1,104 @@
...
+      });
+    });
+  });
+});
+
+function generateSamples() {
+  var samples = [
+    {
+      model: 'purchase',
+      'customer_id': 1,
```
rmg:
```js
var samples = [
  {
    model: 'purchase',
    customer_id: 1,          // <------ customer_id, not 'customer_id'
    basket: ['food', 'drink'],
  },
};
```

README.md
```diff
@@ -531,6 +531,34 @@ Not implemented yet, track it in story https://github.com/strongloop/loopback-co
...
+  // 'server/datasources.json' file
+  var ds = server.datasources.cloudantDB;
+
+
+  ds.on('connected', function() {
+    // 1. Please note `ds.connector.viewDocs()` is the correct way to call it,
+    // NOT `ds.viewDocs()`
+    // 2. This api matches the Cloudant endpoint:
+    // GET /db/_design/<design-doc>/_view/<view-name>
+    ds.connector.viewDocs('design_doc', 'view_name', cb);
```
b-admike: maybe expand `cb` to be inline to show what we expect to see in the callback function like `function (err, views)`?

jannyHou:
@b-admike Changes applied. PTAL again, thanks.

PR #136, opened 2017-05-15
Fix updateOrCreate

ssh24:
### Description

Allow use of `updateAttributes` on the `updateOrCreate` prototype function. Try updating the data using the `updateAttributes` function first and if it fails, create the instance.

connect to https://github.com/strongloop-internal/scrum-apex/issues/204

### DIFF
--- a/lib/cloudant.js
+++ b/lib/cloudant.js
@@ -744,13 +744,15 @@ Cloudant.prototype.updateOrCreate = function(model, data, cb) {
   var idName = self.idName(model);
   var id = data[idName];
   if (id) {
-    self.getCurrentRevision(model, id, function(err, rev) {
-      if (err) return cb(err);
-      if (rev) data._rev = rev;
-      self.create(model, data, {}, function(err) {
-        if (err) return cb(err);
-        cb(err, data, {isNewInstance: !rev});
-      });
+    self.updateAttributes(model, id, data, {}, function(err, docs) {
+      if (err && err.statusCode && err.statusCode === 404) {
+        self.create(model, data, {}, function(err) {
+          if (err) return cb(err);
+          return cb(err, data, {isNewInstance: true});
+        });
+      } else {
+        return cb(err, data, {isNewInstance: false});
+      }
     });
   } else {
     self.create(model, data, {}, function(err, id) {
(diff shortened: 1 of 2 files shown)


lib/cloudant.js
```diff
@@ -710,7 +710,7 @@ Cloudant.prototype.updateAttributes = function(model, id, data, options, cb) {
     if (err) return cb(err);
     data = getPlainJSONData.call(self, model, data);
     _.mergeWith(doc, data, function(dest, src) { return src; });
-    self.create(model, data, options, function(err) {
+    self.create(model, doc, options, function(err) {
```
b-admike: Good catch!

test/cloudant.test.js
```diff
@@ -117,6 +117,64 @@ describe('cloudant connector', function() {
...
+        should.not.exist(err);
+        should.exist(result);
+        result.id.should.equal(newData.id);
+        result.name.should.equal(newData.name);
+        result.age.should.equal(newData.age);
+        done();
+      });
+    });
+
+    it('update existing instance if instance exists', function(done) {
```
b-admike: I would update one of the instances created by the `create` function in L#136

ssh24: I agree. Fixed it.

dhmlau:
Not sure how it's done in other connectors.. For updateOrCreate, would the logic be the following instead: 
- check if the instance exists 
  - if exists, update
  - if not exist, create

ssh24:
@dhmlau That is exactly what we did on this PR. `updateAttributes` throws an error **404** if the instance does not exist, so we check that before creation and if it does not throw an error, the instance has been updated.

dhmlau:
I was thinking if there’s a better function to check the existence of the instance instead of calling the `updateAttributes` to find out

ssh24:
I am not sure why we would want to do another call to a function to check the existence and then call `updateAttributes` later on when calling `updateAttributes` first already does half the job for us. We can catch the error in `updateAttributes` and if it returns a non-existent code we create the instance.

dhmlau:
that's true too.  I was thinking what if the `updateAttribute` function failed for a different reason, the error message we're displaying to the users might be different?  
Anyway, I'm ok either way.  Thanks.

PR #138, opened 2017-05-18
Allow travis to run against the latest code base

ssh24:
Allow travis to run against the latest code base instead of npmjs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ env:
 
 before_install:
   - sudo sh setup.sh
+  - npm config set registry http://ci.strongloop.com:4873/
 
 node_js:
   - 4


.travis.yml
```diff
@@ -11,6 +11,7 @@ env:
 
 before_install:
   - sudo sh setup.sh
+  - npm config set registry http://ci.strongloop.com:4873/
```
rmg: This registry isn't meant to be accessed externally. If this repo depends on unpublished upstream changes, then those changes _need_ to be published.

ssh24: I agree. Our release process is not the fastest. I do not see any other way to get pass this if a repo depends heavily on travis.

PR #139, opened 2017-05-19
test: replace setup.sh with test.js

rmg:
connected to https://github.com/strongloop-internal/scrum-apex/issues/216

Rather than using a shell script use JS and the dockerode module to set
up and teardown a cloudant-developer docker container to test against.

@kjdelisle this is an alternative to #137 that would trigger cis-jenkins to detect that docker-in-docker support is needed.


test.js
```diff
@@ -0,0 +1,195 @@
...
+      }
+      console.log('starting container...');
+      docker.createContainer({
+        Image: 'ibmcom/cloudant-developer:1.0.1',
+        name: 'cloudant-testdb',
+        HostConfig: {
+          PortBindings: {
+            '80/tcp': [{HostPort: '8080'}],
+          },
+          Privileged: true,
```
kjdelisle: I thought this container update no longer required this option?
https://hub.docker.com/r/ibmcom/cloudant-developer/ 
`You no longer need/can run the container in privilege mode.`

test.js
```diff
@@ -0,0 +1,195 @@
...
+
+function runTests(container, next) {
+  var env = JSON.parse(JSON.stringify(process.env));
+  env.CLOUDANT_DATABASE = 'test-db';
+  env.CLOUDANT_PASSWORD = 'pass';
+  env.CLOUDANT_PORT = '8080';
+  env.CLOUDANT_URL = 'http://127.0.0.1:8080';
+  env.CLOUDANT_USERNAME = 'admin';
+  console.log('running mocha...');
+  var mocha = require.resolve('mocha/bin/mocha');
```
kjdelisle: `mocha/bin/_mocha`?

rmg: True. In this case we can call that sub-script directly.

test.js
rmg: That's for `:latest` and `:2.0.0`, but this is still using the `:1.0.1` generation.

rmg:
Ok, @kjdelisle, _now_ I'm done poking at this. PTAL.

test.js
```diff
@@ -0,0 +1,193 @@
...
+}
+
+function run(cmd) {
+  return function spawnNode(container, next) {
+    console.log('running mocha...');
+    spawn(process.execPath, cmd, {stdio: 'inherit'})
+      .on('error', next)
+      .on('exit', onExit);
+
+    function onExit(code, sig) {
```
kjdelisle: We're not using `sig`.

test.js
```diff
@@ -0,0 +1,193 @@
...
+    var opts = {
+      method: 'PUT',
+      path: '/' + db,
+      host: process.env.CLOUDANT_HOST,
+      port: process.env.CLOUDANT_PORT,
+      auth: process.env.CLOUDANT_USERNAME + ':' + process.env.CLOUDANT_PASSWORD,
+    };
+    console.log('creating db: %j', db);
+    http.request(opts, function(res) {
+      res.on('data', function() {});
```
kjdelisle: Did we want to log the data, or should we drop the listener?

test.js
```diff
@@ -0,0 +1,193 @@
...
+    return setImmediate(ping, null, 30);
+
+    function ping(err, tries) {
+      console.log('ping (%d/30)', 30 - tries);
+      if (tries < 1) {
+        next(err || new Error('failed to contact Cloudant'));
+      }
+      http.get(opts, function(res) {
+        res.on('error', tryAgain);
+        res.on('data', function() {});
```
kjdelisle: Did we want to log the data, or should we drop the listener?

test.js
```diff
@@ -0,0 +1,193 @@
...
+  return function waitForPath(container, next) {
+    var opts = {
+      host: process.env.CLOUDANT_HOST,
+      port: process.env.CLOUDANT_PORT,
+      auth: process.env.CLOUDANT_USERNAME + ':' + process.env.CLOUDANT_PASSWORD,
+      path: path,
+    };
+
+    console.log('waiting for instance to respond');
+    return setImmediate(ping, null, 30);
```
kjdelisle: Can we hoist 30 out to a top-level var? Maybe even make it an env var at the CI level?

test.js
```diff
@@ -0,0 +1,193 @@
...
+
+    function ping(err, tries) {
+      console.log('ping (%d/30)', 30 - tries);
```
kjdelisle: Same as above.

package.json
```diff
@@ -16,7 +16,7 @@
   "main": "index.js",
   "scripts": {
     "lint": "eslint .",
-    "test": "mocha -G --timeout 40000",
+    "test": "node test.js mocha -G --timeout 40000",
```
kjdelisle: 2fast5me

test.js
rmg: I'll definitely move it up in scope. Not sure about putting it in CI, though.

test.js
rmg: The no-op `'data'` listener is there to consume the stream so that it ends. I'll try to make that more clear.

test.js
rmg: The no-op `'data'` listener is there to consume the stream so that it ends. I'll try to make that more clear.

rmg:
@strongloop/sq-lb-apex I've addressed the comments in separate fixup commits so you can see how I responded.

test.js
rmg: Added it to the error message.

rmg:
@slnode test please

ssh24:
@slnode test please

PR #140, opened 2017-05-26
Allow users to spawn docker and run tests

ssh24:
Dockerode script is good for CI, but it cleans up after the test
The script keeps the docker container alive and allows to debug easily on test failures

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -1,0 +1,124 @@
+#!/bin/bash		
+ 		
+ # Shell script to start the database and app services before running the tests.		
+ 		
+ ## color codes		
+ RED='\033[1;31m'		
+ GREEN='\033[1;32m'		
+ YELLOW='\033[1;33m'		
+ CYAN='\033[1;36m'		
+ PLAIN='\033[0m'		
+ 		
+ ## variables		
+ HOST=localhost		
+ USER='admin'		
+ PASSWORD='pass'		
+ PORT=8080		
+ DATABASE='testdb'		
+ if [ "$1" ]; then		
+     HOST=$1		
+ fi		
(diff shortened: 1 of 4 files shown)


setup.sh
```diff
@@ -0,0 +1,119 @@
...
+     PASSWORD=$4		
+ fi		
+ if [ "$5" ]; then		
+     DATABASE=$5		
+ fi		
+ 		
+ ## check if docker exists		
+ printf "\n${RED}>> Checking for docker${PLAIN} ${GREEN}...${PLAIN}"		
+ docker -v > /dev/null 2>&1		
+ DOCKER_EXISTS=$?
```
b-admike: Check for `docker-compose` as well since we use it in this script or remove the dependency on it and this can be the same.

ssh24: Good catch. Fixed.

setup.sh
```diff
@@ -0,0 +1,119 @@
...
+     PASSWORD=$4		
+ fi		
+ if [ "$5" ]; then		
+     DATABASE=$5		
+ fi		
+ 		
+ ## check if docker exists		
+ printf "\n${RED}>> Checking for docker${PLAIN} ${GREEN}...${PLAIN}"		
+ docker -v > /dev/null 2>&1
+ docker-compose -v > /dev/null 2>&1
```
b-admike: If `$DOCKER_EXISTS` checks for both the output of docker and docker-compose checks then this is good. If not, I suggest you have another var `$DOCKER_COMPOSE_EXISTS` for it separately and check for both below

ssh24: Fixed.

Issue #35, opened 2016-09-27
Fix update/updateOrCreate in cloudant.js 


jannyHou:
Use `updateAttributes` function to `update` a model instance and use `update` function in call for `updateOrCreate`. 

Currently, `update` performs a destructive update. With this change `update` function should be partial update.

djorg83:
Possibly related to #110

jannyHou:
@djorg83 thanks for catching the error in #110 , this is not the same one.
cloudant only support full replace of document, but not partial update.
This issue will implement an update method which does partial update.

jannyHou:
Some possible solutions:

- Keep it same as `replaceOrCreate`, and warn users there is not partial update.
- (Tricky) We implement a way to merge data from document and updated data, then call `_insert` , which will do replace
- Throw error/warning in this method to explicitly tell user they need to define their own method if they do want to partial update.

cwadding:
@jannyHou one solution for a partial update are to use update handlers https://cloud.ibm.com/docs/services/Cloudant/api?topic=cloudant-design-documents#update-handlers
This would need to be enabled on the cloudant database but can make it so that the updates are not a destructive operation.
It might would be nice to be able to specify the desired update handler

PR #141, opened 2017-05-29
Fix updateAttributes method

ssh24:
### Description

~Currently in cloudant `update` performs a destructive update. With this change `update` function should do partial update.~

~- Use `updateAttributes` function to `update` a model instance.~
~- Use `update` function in `updateOrCreate` to update a model instance if exists.~

- Fix `updateAttributes` function to return the docs from `fromDB` function correctly. This is required since `updateOrCreate` calls `updateAttributes` now.
- Remove duplicate `updateAll` tests
- Allow `fromDB` function to return the id as a number if the original property was a number

connect to https://github.com/strongloop/loopback-connector-cloudant/issues/35

### DIFF
--- a/lib/cloudant.js
+++ b/lib/cloudant.js
@@ -134,10 +134,16 @@ Cloudant.prototype.toDB = function(modelName, modelObject, doc) {
  */
 Cloudant.prototype.fromDB = function(modelName, modelObject, doc) {
   var idName = this.idName(modelName);
+  // we should return the `id` as an int if the user specified the property as an int
+  if (idName)
+    var idType = modelObject.mo.properties.id.type.name;
   if (!doc) return doc;
   assert(doc._id);
   if (doc._id) {
-    doc[idName] = doc._id.toString();
+    if (idType === 'Number')
+      doc[idName] = parseInt(doc._id);
+    else
+      doc[idName] = doc._id;
     delete doc._id;
   }
   for (var i = 0; i < modelObject.dateFields.length; i++) {
@@ -463,7 +469,7 @@ Cloudant.prototype.updateAttributes = function(model, id, data, options, cb) {
     self.create(model, doc, options, function(err, id, rev) {
       if (err) return cb(err);
       doc._rev = rev;
-      return cb(err, doc);
+      return cb(err, self.fromDB(model, mo, doc));
     });
   });
 };
@@ -500,7 +506,7 @@ Cloudant.prototype.updateOrCreate = function(model, data, cb) {
         delete data._rev;
         self.create(model, data, {}, createHandler);
       } else {
-        return cb(err, self.fromDB(model, mo, docs), {isNewInstance: false});
+        return cb(err, docs, {isNewInstance: false});
       }
     });
   } else {
--- a/test/cloudant.test.js
+++ b/test/cloudant.test.js
@@ -70,41 +70,6 @@ describe('cloudant connector', function() {
     });
   });
 
-  describe('updateAll and updateAttributes', function() {
-    var productInstance;
-    beforeEach('create Product', function(done) {
-      Product.create({
-        id: 1,
-        name: 'bread',
-        price: 100,
-      }, function(err, product) {
-        if (err) return done(err);
-        productInstance = product;
-        done();
-      });
-    });
-
-    afterEach(function(done) {
-      Product.destroyById(1, function(err) {
-        if (err) return done(err);
-        done();
-      });
-    });
-
-    it('accepts loopback model instance as input data for update',
-      function(done) {
-        productInstance.setAttribute('name', 'butter');
-        Product.updateAll({id: '1'}, productInstance, function(err, res) {
-          if (err) return done(err);
-
-          Product.findById('1', function(err, prod) {
-            prod.name.should.equal('butter');
-            done();
-          });
-        });
-      });
-  });
-
   describe('model with array props gets updated properly', function() {
     var prod1, prod2;
     before('create Product', function(done) {
--- a/test/update.test.js
+++ b/test/update.test.js
@@ -154,3 +154,50 @@ describe('updateAll', function() {
     });
   });
 });
+
+describe('updateAttributes', function() {
+  before(function(done) {
+    db = getDataSource();
+
+    Product = db.define('Product', {
+      name: {type: String},
+      description: {type: String},
+      price: {type: Number},
+    }, {forceId: false, updateOnLoad: true});
+
+    db.once('connected', function() {
+      db.automigrate(function(err) {
+        Product.create(bread, done);
+      });
+    });
+  });
+
+  after(cleanUpData);
+
+  it('update an attribute for a model instance', function(done) {
+    var updateFields = {
+      name: 'bread2',
+    };
+    Product.find(function(err, result) {
+      testUtil.hasError(err, result).should.not.be.ok();
+      testUtil.hasResult(err, result).should.be.ok();
+      var id = result[0].id;
+      var oldRev = result[0]._rev;
+      var newData = _.cloneDeep(result[0]);
+      newData.name = updateFields.name;
+      var product = new Product(result[0]);
+      product.updateAttributes(newData, function(err, result) {
+        testUtil.hasError(err, result).should.not.be.ok();
+        testUtil.hasResult(err, result).should.be.ok();
+        var newRev = result._rev;
+        oldRev.should.not.equal(newRev);
+        Product.find(function(err, result) {
+          testUtil.hasError(err, result).should.not.be.ok();
+          testUtil.hasResult(err, result).should.be.ok();
+          newRev.should.equal(result[0]._rev);
+          done();
+        });
+      });
+    });
+  });
+});


dhmlau:
If the `Partial Update` section in the `README.md` is no longer accurate, we'll need to update that too. https://github.com/strongloop/loopback-connector-cloudant#partial-update

test/cloudant.test.js
```diff
@@ -70,41 +70,6 @@ describe('cloudant connector', function() {
     });
   });
 
-  describe('updateAll and updateAttributes', function() {
```
rmg: Why are these tests being removed? If the functionality they are testing no longer works with these tests, doesn't that mean this is a breaking change?

test/update.test.js
```diff
@@ -144,7 +144,7 @@ describe('updateAll', function() {
       Product.update({id: id}, newData, function(err, result) {
         testUtil.hasError(err, result).should.not.be.ok();
         testUtil.hasResult(err, result).should.be.ok();
-        var newRev = result[0].rev;
```
rmg: If `.rev` was previously correct but now it is not, doesn't that make this a breaking change?

test/cloudant.test.js
ssh24: This test accomplishes the same suite of test in [`update.test.js`](https://github.com/strongloop/loopback-connector-cloudant/blob/master/test/update.test.js#L115)

test/update.test.js
ssh24: That is because `update` didn't work the correct way before. It would not convert the values back from the database correctly.

test/cloudant.test.js
kjdelisle: If the `updateAll and updateAttributes` test was a duplicate of the `updateAll` tests, then that definitely needs fixing, but I would rewrite it to be an isolated test of `updateAttributes` exclusively, rather than deleting it.

kjdelisle:
@rmg The desired behaviour is this:
- `update/updateAll` should use the `updateAttributes` function to perform the semi-destructive updates that we're looking for
- `updateOrCreate` should use the `update` function rather than calling `updateAttributes` directly, which should simplify the logic of that call.
- Test coverage should now hit on each of those three sections individually to ensure that they all perform as expected.

lib/cloudant.js
```diff
@@ -493,14 +499,14 @@ Cloudant.prototype.updateOrCreate = function(model, data, cb) {
   };
 
   if (id) {
-    self.updateAttributes(model, id, data, {}, function(err, docs) {
+    self.update(model, {id: id}, data, {}, function(err, docs) {
       if (err && err.statusCode !== 404) return cb(err);
-      else if (err && err.statusCode === 404) {
+      else if (err && err.statusCode === 404 || docs.count === 0) {
```
kjdelisle: Is that even possible?
If the document has been deleted, or never existed to begin with, I'd expect a `404 Not Found` instead of a docs object with a count of zero.

Additionally, is it `count`, or should that be `length`?

kjdelisle: I'd like to see some test coverage that demonstrates that `docs.count` can be zero, otherwise I'd remove this.

ssh24:
@rmg @kjdelisle Feedback applied. PTAL

Note `updateAttributes` test will fail because juggler needed a fix to allow pass through of the new data. PR: https://github.com/strongloop/loopback-datasource-juggler/pull/1393

lib/cloudant.js
```diff
@@ -524,20 +530,37 @@ Cloudant.prototype.updateAll = function(model, where, data, options, cb) {
...
+      return cb(err);
+    }
 
-    data = getPlainJSONData.call(self, model, data);
     async.each(docs, function(doc, cb) {
+      if (data._rev) {
+        if (doc._rev !== data._rev) {
+          err = {};
+          err.statusCode = 409;
+          err.message = g.f('Document update conflict with rev %s', doc._rev);
```
kjdelisle: Returning an error via `async.each` would mean that the user/app making these calls might have to repeat this a ridiculous number of times to discover all of the individual document conflicts, since the first failure will halt the remainder of these calls.

I would expect that before you attempt the asynchronous `updateAttributes` call, that you would compare the `_.rev` in a loop.
```
var conflicts = [];
_.each(docs, function(doc) {
  var item = _.find(data, { id: doc.id });
  if (item._rev !== data._rev) {
    conflicts.push(doc.id);
  }
});

if (!_.isEmpty(conflicts)) {
  err = new Error(util.format('Document conflict(s):%s%s', os.EOL, conflicts.join(os.EOL));
  err.statusCode = 409;
  return cb(err);
}
// Continue on with the updateAttributes calls...
```

ssh24: That is very true. Thank you for the catch. Fixed.

test/update.test.js
```diff
@@ -144,7 +144,55 @@ describe('updateAll', function() {
...
+      price: 100,
+    };
+    Product.find(function(err, result) {
+      testUtil.hasError(err, result).should.not.be.ok();
+      testUtil.hasResult(err, result).should.be.ok();
+      var id = result[0].id;
+      var oldRev = result[0]._rev;
+      var oldResult = _.cloneDeep(result[0]);
+      oldResult.name = newData.name;
+      var instance = new Product(result[0]);
```
jannyHou: maybe `product` is bette than `instance`. 
`var product = new Product(result[0]); `
and `oldResult` is a little misleading, it actually means the new data to be updated.

```js
var updateFields = {
  name: 'bread2',
  price: 100,
}
var newData = _.cloneDeep(result[0]);
updateFields.forEach(updateFields, function(value, key) {
  newData[key] = value;  
});

produce.updateAttributes(updateFields, cb);
```

rmg:
Sorry, this isn't directly related to this PR.

---
The complicated handling of document conflicts described by @kjdelisle is the kind of thing that always confused me about high-level CouchDB libraries. At the low-level, it is just basic HTTP operations, which is pretty easy to handle with builtin http support in most languages, or with some thing wrapping to make the to/from JSON more automatic. Once you go up a level from that you have to deal with document conflicts and that is inherently app-specific, so it ends up having to be implemented in the app rather than the connector/client anyway.

This is a general problem with generic and multi-db ORMs, though.
 - You have a generic ORM supporting different types of DBs.
 - You can't support all the special feature of each db, so you implement none of them.
 - You aren't making use of a database's special features so you probably aren't doing anything interesting with the DB.
 - You aren't doing anything interesting with the DB so it doesn't matter which DB you use.
 - You use the DB that the ORM supports best.
 - Your ORM drops support for unused DB types.
 - Your ORM only supports a single DB type so it is free to implement that DB's special features.
 - Now your ORM is no longer generic.

lib/cloudant.js
```diff
@@ -524,20 +531,40 @@ Cloudant.prototype.updateAll = function(model, where, data, options, cb) {
...
   var self = this;
   var mo = self.selectModel(model);
+  var resultArray = [];
   self.all(model, {where: where}, {raw: true}, function(err, docs) {
     if (err) return cb(err, docs);
-    if (docs.length === 0) return cb(null, {count: 0});
+    var conflicts = [];
+    if (docs.length === 0) {
+      err = new Error(util.format('No document found with filter: %j%s%s',
+        where, os.EOL, conflicts.join(os.EOL)));
```
b-admike: Is it a formatting necessity or does logging `os.EOL` have any benefits?

ssh24: Fixed.

lib/cloudant.js
```diff
@@ -134,10 +135,16 @@ Cloudant.prototype.toDB = function(modelName, modelObject, doc) {
...
+  if (idName)
+    var idType = modelObject.mo.properties.id.type.name;
   if (!doc) return doc;
   assert(doc._id);
   if (doc._id) {
-    doc[idName] = doc._id.toString();
+    if (idType === 'Number')
+      doc[idName] = parseInt(doc._id);
+    else
+      doc[idName] = doc._id.toString();
```
jannyHou: nitpick, theoretically `doc._id` here is already a `String`

lib/cloudant.js
```diff
@@ -493,14 +500,14 @@ Cloudant.prototype.updateOrCreate = function(model, data, cb) {
...
-    self.updateAttributes(model, id, data, {}, function(err, docs) {
+    self.update(model, {id: id}, data, {}, function(err, docs) {
       if (err && err.statusCode !== 404) return cb(err);
       else if (err && err.statusCode === 404) {
         // strip out the `_rev` property before creation
         delete data._rev;
         self.create(model, data, {}, createHandler);
       } else {
-        return cb(err, self.fromDB(model, mo, docs), {isNewInstance: false});
+        return cb(err, docs[0], {isNewInstance: false});
```
jannyHou: @ssh24 there is a question for [Line#514](https://github.com/strongloop/loopback-connector-cloudant/pull/141/files#diff-fa06ecf3249c63aabcbf7ff1ae8ec186R514), not this line:

IMO, If a user doesn't provide `id` but the data contains `_rev`, then cloudant should refuse to insert the doc, not we "strip out the `_rev` property before creation" instead.

lib/cloudant.js
```diff
@@ -524,20 +531,40 @@ Cloudant.prototype.updateAll = function(model, where, data, options, cb) {
...
+    var conflicts = [];
+    if (docs.length === 0) {
+      err = new Error(g.f('No document found with filter: %j%s%s',
+        where, os.EOL, conflicts.join(os.EOL)));
+      err.statusCode = 404;
+      return cb(err);
+    }
+    _.each(docs, function(doc) {
+      if (data._rev) {
+        if (data._rev !== doc._rev) conflicts.push(doc._id);
```
jannyHou: There is a bulk update method provided by the cloudant driver, I think we should use that method to `updateAll`.
> `if (data._rev !== doc._rev) conflicts.push(doc._id);`

would not be a valid check for an array of instances, each instance has its own `_rev`, we could not expect user provide a single `_rev` that could update all instances....

Haven't figured out what is the proper way to do a bulk update with `_rev` provided in data, I will double check juggler code :(

lib/cloudant.js
ssh24: We do need to strip the `_rev` before creation though. If the user tries to do a create and pass in a random `_rev` value, it is upon us to strip it off before insertion.

This is not related to this PR at all. It was well discussed and accepted for the PR of [`allow _rev property`](https://github.com/strongloop/loopback-connector-cloudant/pull/123)

lib/cloudant.js
ssh24: @jannyHou You are kind of correct. The condition I had there was wrong. Fixed it to account for both `id` and `_rev`. Tests should have been failing before. Should pass now (except the `updateAttributes` which still fails cause of juggler).

Doing bulk update directly was the way we previously did it. As discussed per the squad, we want to write the update functions so they work in this way:
- Use `updateAttributes` function to `update` a model instance.
- Use `update` function in `updateOrCreate` to update a model instance if exists.

lib/cloudant.js
kjdelisle: We could handle it either by stripping off `_rev` or by throwing an error if one is provided. In either case, we'll have to do _something_ to stop the user.

jannyHou: sorry I was not involved in the whole discussion of pr https://github.com/strongloop/loopback-connector-cloudant/pull/123

> by stripping off `_rev`

Then cloudant will just insert the data right? It won't throw error for a data without `_id` and `_rev`

ssh24: Ok, I think I understand now. On the previous PR #123, we just stripped of `_rev` instead of throwing an error. Talked with @jannyHou and we came to a conclusion of throwing an error is the better way to go if the user specifies `_rev` on a create operation.

We shall create another story/PR for it and keep it separate from this one.

jannyHou:
```js
Test.updateAll({age: 1}, {age: 2}, cb);  //WRONG
``` 
would not be a valid scenario considering we have introduced the `_rev` property: 
If you DON NOT provide `_rev`, you CAN NOT do update, you only get a 409 conflict error.

A valid scenario would be:
```js
Test.find({age:1}).
  then(function(results) {
    var newData = {age: 2};
    var newInstances = _.map(results, function(r) {
      newData.forEach(function(prop){r[prop] = newData[prop]});
      return r;  
    });
    return newInstances;
  }).then(function(newInst) {
      Test.updateAll(newInst);
  });
```

jannyHou:
@ssh24 Continue our previous discussion: 
IMO, 
```js
Test.updateAll({age: 1}, {age: 2}, cb);  //WRONG
``` 
would not be a valid scenario considering we have introduced the `_rev` property: 
If you DON NOT provide `_rev`, you CAN NOT do update, you only get a 409 conflict error.

A valid scenario would be:
```js
Test.find({age:1}).
  then(function(results) {
    var newData = {age: 2};
    var newInstances = _.map(results, function(r) {
      return _.extend(r, newData);  
    });
    return newInstances;
  }).then(function(newInsts) {
      // if {data} has to be an object, then wrap the array with some property // like `updateData`
      // if {data} could be either object or array, cool, in cloudant function
      // just pass `newInsts` into bulkupdate
      Test.updateAll({}, {updateData: newInsts});
  });

Cloudant.prototype.updateAll(model, where, data, cb) {
    // just ignore where filter
    Cloudant.bulkupdate(data.updateData);
}
```

ssh24:
@jannyHou I am still unconvinced with what you are saying.

Let's say I have three instances of a model in the database:
```javascript
[{id: 1, name: 'Foo', age: 1},
{id: 2, name: 'Bar', age: 1},
{id: 3, name: 'Baz', age: 2}]
```

Now in the database itself, it is stored with **unique** _rev value for the three instances.

The user can then do `updateAll` on a model instance. Let's say the user wanted to **updateAll** model instances where age = 1. Why do we care about _rev here until and unless they want to filter with the _rev value as well in which case it should work the same way as well?

When the user filters by {age: 1}, it gets back the two instances:
```javascript
[{id: 1, name: 'Foo', age: 1},
{id: 2, name: 'Bar', age: 1}]
```

And then the connector updates those two instances with the data provided.

Now, as long as the user did not specify the _rev value in the data provided to be updated, we are good. If the user did provide a _rev value in the data to be updated and it does not match with the one in the database, we throw 409 error.

Let's discuss about it more tomorrow in person. :)

ssh24:
@slnode test please

jannyHou:
@ssh24 The original purpose of that `_rev` PR is to stop handling the `_rev` in connector.
 
> Why do we care about _rev here until and unless they want to filter with the _rev value as well in which case it should work the same way as well?

I am not sure I understand this. Filter is not a good reason to render the connector to fetch instances, then use their `_rev` to do the update without giving users a chance to handle them...

jannyHou:
@ssh24 I was not aware of loopback has an api called `bulkUpdate`, what I put in comment  https://github.com/strongloop/loopback-connector-cloudant/pull/141#issuecomment-305050368 should actually be the implementation of `Cloudant.prototype.bulkUpdate`.

If updateAll with a filter has to fetch instance and use its `_rev`, I have no idea how to avoid handling  `_rev` in connector.

kjdelisle:
Honestly, if we have `bulkUpdate`, then I'd expect `updateAll` to work the LoopBack way, but I would also expect it to fail a lot on large volumes of data/high traffic data. Additionally, due to the way it's written, a lot of the updates would succeed even though we're returning failures because of conflicts.

The idea of `updateAll` almost assumes a transactional lock that would safely allow that to happen, and revert it if it couldn't.

Personally, I'd break the `update` and `updateAll` functions into two separate ones, and then make `updateAll` throw an exception. 

@jannyHou @ssh24 Is that something we can do, or would that violate some expectations in Juggler/DAO?

rmg:
I think there needs to be a discussion around the direction of the connector. It should either try to completely _hide_ `_rev` and try to handle it all automatically (and probably have poor performance as a result), or it should expose `_rev` so that the user can properly work with conflicts in which case it doesn't really behave like a LoopBack connector.

It might also be worth talking to the maintainers of [loopback-connector-couchdb](https://github.com/Wiredcraft/loopback-connector-couchdb) to see how they feel, since they would have the same problems.

kjdelisle:
@rmg I think we've already decided that our intent is to expose `_rev` as much as we can and not shelter Cloudant users from the necessity of handling it, or building application logic around it.

At this point, I think Juggler/DAO's tolerance for this should only be a factor when it breaks a use-case entirely.

@jannyHou Reaching out to the developers of that connector would definitely be a good idea, as well as giving their modules a more thorough reading. I'd also make a test app with that connector so you can watch it work from a different perspective.

kjdelisle:
^^^ cc @ssh24

ssh24:
@kjdelisle `update` and `updateAll` are aliases in Juggler. So I think your idea of breaking them into two different functions might not work as expected.

jannyHou:
@ssh24 If here we need a function that can **do a single update**, then I think you can use `Cloudant.prototype._insert` right?

ssh24:
> If here we need a function that can do a single update, then I think you can use Cloudant.prototype._insert right?

@jannyHou That is just doing a `create`? How will that work?

jannyHou:
@ssh24 I am not sure what's the original purpose of calling an `update` method, but all `create` and `update` related methods are calling the same cloudant api: `_insert`.

test/cloudant.test.js
```diff
@@ -102,61 +102,73 @@ describe('cloudant connector', function() {
...
+      Product.updateAll({id: 1}, prod1, function(err, res) {
+        if (err) return done(err);
+        Product.findById('1', function(err, res) {
+          if (err) done(err);
+          res.name.should.equal(prod1.name);
+          res.price.should.equal(prod1.price);
+          res.releases.should.deepEqual([4, 5, 6]);
+          res.type.should.deepEqual(['cinnamon raisin']);
+          res.foo.should.deepEqual([{id: 3, name: 'bread3'}]);
+          Product.findById('2', function(err, res) {
```
jannyHou: any reason why we verify a instance not get updated?

test/cloudant.test.js
```diff
@@ -102,61 +102,73 @@ describe('cloudant connector', function() {
...
+    it('updates all matching instances with array props', function(done) {
+      var data = {
+        price: 200,
+        releases: [7],
+        type: ['everything'],
+        foo: [{id: 1, name: 'bar'}],
+      };
+      Product.find(function(err, result) {
+        if (err) done(err);
+        var id1 = result[0].id;
```
jannyHou: The usage of `result` here is a little dangerous...how do we make sure there are two items in the result :)

maybe `result.should.be.an.Array();` and `result.length.should.equal(2)`

test/cloudant.test.js
ssh24: No idea :\ It was always there [testing both instances](https://github.com/strongloop/loopback-connector-cloudant/blob/master/test/cloudant.test.js#L154) :p

test/cloudant.test.js
```diff
@@ -102,61 +102,73 @@ describe('cloudant connector', function() {
...
+        var id2 = result[1].id;
+        var rev1 = result[0]._rev;
+        var rev2 = result[1]._rev;
+        data.id = id1;
+        data._rev = rev1;
+
+        Product.updateAll({price: 100}, data, function(err, res) {
+          if (err) done(err);
```
jannyHou: what is the purpose of doing a `updateAll` here? Sorry I am confused with the two `updateAll` called in this test, the logic seems like this to me:

- do a `updateAll` (which I am not sure what is the purpose)
- modify the `data`(which will be the new data) to match the second instance's `_id` and `_rev`
- do another `updateAll`, then the first and second instances both get updated....?

Could you elaborate more lol?

ssh24: You are partially correct. The way `updateAll` works now is basically an **update** on a single instance because it takes into account the `_rev` and `_id` property (which are always unique!).

 The test does this:
- First does a `find` and stores the four values
```
        var id1 = result[0].id;
        var id2 = result[1].id;
        var rev1 = result[0]._rev;
        var rev2 = result[1]._rev;
```
- Then appends, `id1` and `rev1` to the data values to be updated and calls `updateAll`
- After that update is complete, it adds `id2` and `rev2` to the data and calls `updateAll`
(Just to be clear both those id and rev values are added as `id` and `_rev` property in the data to be updated)
- Finally, it just does a find to check if both the instances got updated.

test/cloudant.test.js
ssh24: Good catch. Fixed :)

lib/cloudant.js
```diff
@@ -524,21 +531,52 @@ Cloudant.prototype.updateAll = function(model, where, data, options, cb) {
...
+
   self.all(model, {where: where}, {raw: true}, function(err, docs) {
     if (err) return cb(err, docs);
     if (docs.length === 0) return cb(null, {count: 0});
 
     data = getPlainJSONData.call(self, model, data);
+
+    /**
+     * for each of the docs from the given filter, only update the instance where the revs match
+     * behavior of bulk update in cloudant:
```
kjdelisle: If this comment is reflective of the code below, then it's not what we talked about.
We want to apply the bulk update to all of the documents, regardless of what they've done; we're just going to remove `_rev` from the `docs` entries so that we don't force-overwrite items whose `_rev` values might be newer/different.

By doing that, we should be covered for these scenarios:
`_rev` is missing from one or more objects in `data`: They'll fail the rev-check in Cloudant, and will be aggregated in the list of failures returned to us in the results payload.
`_rev` is on all of the entries in `data`: Cloudant will still aggregate any conflicting documents together, and then we'll be able to return an error that includes the items that failed.

The downside to this is that we have to return an error code even though some of the bulk update succeeded (which is in contrast with Cloudant's return of a 201 or 202 depending on *how* successful your bulk update was). 
Our hands are tied in this regard due to the LoopBack API's expectation that updateAll returns a record *count*, and not a full list of items updated.

**TL;DR** - Remove this comment, make sure the code isn't doing what this comment says it's doing.

lib/cloudant.js
```diff
@@ -524,21 +531,52 @@ Cloudant.prototype.updateAll = function(model, where, data, options, cb) {
...
+       - if both _id and _rev are not provided, return empty list
+       - if _id is provided and _rev is not provided, throw error
+       - if _id is not provided and _rev is provided, do update
+       - if both _id and _rev are provided, do update
+    **/
     async.each(docs, function(doc, cb) {
-      _.mergeWith(doc, data, function(dest, src) { return src; });
+      if (doc._id === data._id || doc._rev === data._rev) {
+        delete doc._id;
```
kjdelisle: Not necessary, and frankly, not correct.
We want to make sure that what we're merging is being merged correctly.

lib/cloudant.js
```diff
@@ -524,21 +531,52 @@ Cloudant.prototype.updateAll = function(model, where, data, options, cb) {
...
+        delete doc._rev;
+        _.mergeWith(doc, data, function(dest, src) { return src; });
```
kjdelisle: Given the comment above, are we certain that `_.mergeWith` will be merging the correct items in each array together, or does it only line them up by index?

You might want a test to validate that the correct objects are being combined within the arrays _in the correct way_ (eg. the `docs` array's version of `{id: 1}` is being merged with the `data` array's version of `{id: 1}`)

kjdelisle: The current tests are pre-ordering both collections, which would hide this issue (if it is an issue).

lib/cloudant.js
ssh24: I do not quite understand what you mean.

> _rev is missing from one or more objects in data: They'll fail the rev-check in Cloudant, and will be aggregated in the list of failures returned to us in the results payload.

data is only one object to be updated. when the user specifies the `_rev`, bulk update will at most do one update to the doc where the docs match since `_rev` is **unique**.

The use case is simple:
```javascript
      Employee.create([
        {id: 1, name: 'Foo', age: 1},
        {id: 2, name: 'Bar', age: 1},
        {id: 3, name: 'Baz', age: 2},
      ],
      function(err, result) {
        if (err) throw err;
        console.log('\nCreated instance: ' + JSON.stringify(result));
        var  idInst1 = result[0].id;
        var revInst1 = result[0]._rev;

        Employee.update({age: 1}, {name: 'Sakib', age: 12, _rev: revInst1, id: idInst1},
        function(err, result) {
          if (err) throw err;
          console.log('\nUpdated instance: ' + JSON.stringify(result));

          Employee.find(function(err, result) {
            if (err) throw err;

            console.log('\nFound instance: ' + JSON.stringify(result));
          });
        });
      });
```

Feel free to explain your concern with some code.

rmg: I'm not sure what the uniqueness of `_rev` has to do with anything here. It is a per-document number with a digest appended to it so that you know that `2-deadbeef` is newer than `1-abcdef`, but you can't guess/forge what `3-??????` will be and accidentally overwrite it.

kjdelisle:
```
// Change all the names!
var data = [
 {_id: '1', '_rev':'1-123456', name: 'Sakib'},
 {_id: '2', '_rev': '1-432423', name: 'Janny'},
 {_id: '3', '_rev': '1-612412', name: 'Kevin'},
];

var docs = [
 {_id: '1', '_rev':'1-123456', name: 'Bob'},
 {_id: '2', '_rev': '1-432423', name: 'Ted'},
 {_id: '3', '_rev': '1-612412', name: 'Joe'},
];

_.shuffle(docs); // Maybe the docs aren't in the same order when we get them!

_.mergeWith(docs, data, function(dest, src) { return src; });
```

When we run the update, are we sure that the values from data are being merged together with their corresponding items in docs?
If we're deleting `_id` out of them, how do we match them up to do the merge? Even if we don't do that, where's the logic that's using the `_id` field to determine which objects to zip together?

ssh24: First things first, `data` passed to the `update` function is an **object** not an **array of object**.

Being said that, the code I have right now deletes the `_id` and `_rev` out of each of the docs we get back **iff** `doc._id === data._id || doc._rev === data._rev` and then merges the data with the doc. 

Upon that condition, it also adds to the an array of docLists that needs to be updated and on the final callback it calls `mo.db.bulk({docs: docList}, function(err, result) {...});`

On the callback, we check for the result and if there are any elements in the result array that has `error`, we throw the error. Else we just return the count of the number of result updated.

kjdelisle: Disregard the previous deleted comments; I was thinking of `updateAll` as loopback's `bulkUpdate`.

Issue #37, opened 2016-10-04
updateAll function should return an object


jmereaux:
Hi,

According to this [link](https://github.com/strongloop/loopback/issues/1167), the updateAll function of the connector should now return an object like:

`{ count: <numberOfRecordsUpdated> }` 

Could you please make the change?
Thanks

jannyHou:
Hi @jmereaux Thanks for reminding us of making changes consistently.

If I understand correctly, now [`Cloudant.prototype.updateAll`](https://github.com/strongloop/loopback-connector-cloudant/blob/master/lib/cloudant.js#L563) returns the whole result object instead of the count only.

I reproduced it on my local. Label it as a bug here.

ssh24:
Issue fixed on PR https://github.com/strongloop/loopback-connector-cloudant/pull/145.

Also backported for 1.x https://github.com/strongloop/loopback-connector-cloudant/pull/146

1.x shall be released first, while the fix on 2.x will go out when 2.x is released in the upcoming weeks.

PR #145, opened 2017-05-30
Make updateAll return count of updated instances

ssh24:
Currently, cloudant returns an array of the updated instances instead of the count of the instances updated. This behaviour is not consistent with other connectors. The fix returns the number of instances updated instead of the instances itself.

connect to https://github.com/strongloop/loopback-connector-cloudant/issues/37

### DIFF
--- a/lib/cloudant.js
+++ b/lib/cloudant.js
@@ -534,9 +534,11 @@ Cloudant.prototype.updateAll = function(model, where, data, options, cb) {
       return cb();
     }, function(err) {
       if (err) return cb(err);
-      mo.db.bulk({docs: docs}, function(err, result) {
-        if (result) result.count = docs.length;
-        return cb(err, result);
+      mo.db.bulk({docs: docs}, function(err) {
+        if (err) return cb(err);
+        var res = {};
+        res.count = docs.length;
+        return cb(err, res);
       });
     });
   });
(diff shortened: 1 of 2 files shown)


kjdelisle:
What's the purpose of this PR given the goals of #141 ?
This whole section of code you're changing will be removed as a result of that PR, should it land.

ssh24:
Not really. PR https://github.com/strongloop/loopback-connector-cloudant/pull/141 still returns the objects that were updated. We need to return the count of the objects instead as per the issue related to it. There will be merge conflicts, but I will deal with them.

kjdelisle:
I mostly meant that you can just make those fixes in the other PR and not worry about doing it here.

Did we intend to backport this fix to 1.x? If so, then disregard as that would make sense.

ssh24:
That is what I thought at first well. But since they are two different issues (like github based stories), decided to do them in separate PR's.

Also I think this PR should be backported as the current behaviour is not consistent. Might as well make 1.x consistent as well.

lib/cloudant.js
```diff
@@ -534,9 +534,11 @@ Cloudant.prototype.updateAll = function(model, where, data, options, cb) {
       return cb();
     }, function(err) {
       if (err) return cb(err);
-      mo.db.bulk({docs: docs}, function(err, result) {
-        if (result) result.count = docs.length;
-        return cb(err, result);
+      mo.db.bulk({docs: docs}, function(err) {
+        if (err) return cb(err);
+        var res = {};
```
kjdelisle: We don't want to return the data anymore?

ssh24: Nope. All other connector just returns the count of the number of instances updated, so this fix is to make cloudant consistent with the rest.

PR #146, opened 2017-05-30
Backport: return back result count in updateAll

ssh24:
Backport for #145 

connect to https://github.com/strongloop/loopback-connector-cloudant/issues/37

### DIFF
--- a/test/update.test.js
+++ b/test/update.test.js
@@ -58,7 +58,8 @@ describe('updateAll', function() {
           should.not.exist(err);
           should.exist(result);
           newData.id = id;
-          should.deepEqual(newData, result[0].__data);
+          should.equal(newData.name, result[0].name);
+          should.equal(newData.price, result[0].price);
           done();
         });
       });
--- a/test/update.test.js
+++ b/test/update.test.js
@@ -1,0 +1,67 @@
+// Copyright IBM Corp. 2016. All Rights Reserved.
+// Node module: loopback-connector-cloudant
+// This file is licensed under the Artistic License 2.0.
+// License text available at https://opensource.org/licenses/Artistic-2.0
+
+'use strict';
+
+require('./init.js');
+var Cloudant = require('../lib/cloudant');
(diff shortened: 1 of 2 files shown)


test/update.test.js
```diff
@@ -0,0 +1,67 @@
...
+      Product.update({id: id}, newData, function(err, result) {
+        should.not.exist(err);
+        should.exist(result);
+        result.should.have.property('count');
+        result.count.should.equal(1);
+        Product.find(function(err, result) {
+          should.not.exist(err);
+          should.exist(result);
+          newData.id = id;
+          should.deepEqual(newData, result[0].__data);
```
b-admike: `result[0].data`

test/update.test.js
```diff
@@ -0,0 +1,67 @@
+// Copyright IBM Corp. 2016. All Rights Reserved.
+// Node module: loopback-connector-cloudant
+// This file is licensed under the Artistic License 2.0.
+// License text available at https://opensource.org/licenses/Artistic-2.0
+
+'use strict';
+
+require('./init.js');
+var Cloudant = require('../lib/cloudant');
```
b-admike: Nitpick: sort required modules

ssh24: They are sorted correctly, though.

kjdelisle: Capital letters first, then `_`, then lowercase letters (it's sorted correctly).

When in doubt: `:'<,'>sort`

test/update.test.js
kjdelisle: Nope, that's correct as-is. `__data` is the underlying structure that holds the values. There shouldn't be a corresponding public property.

kjdelisle: We override the getter/setters on the objects so that they modify a `__data` collection under the hood for each of the properties defined in the Model objects. So if you've got a `foo` and `bar` property in your model, they're each configured to get/set from `__.data[foo]` and `__data[bar]` respectively.

PR #149, opened 2017-06-02
Check error and result

jannyHou:
### Description

Callback error message in tests.


jannyHou:
Thanks @kjdelisle , I will merge this first then change to your style in the next PR, I believe there will be more PRs in the near future

PR #150, opened 2017-06-06
Add cloudant specific bulkReplace function

ssh24:
### Description

Implement a cloudant specific bulk update function. Call it `massUpdate`. The function takes in an array of data and performs bulk update against the model instance where the `_rev` of the datas match.

connect to https://github.com/strongloop/loopback-connector-cloudant/issues/35

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,7 @@ The `loopback-connector-cloudant` module is the Cloudant connector for the LoopB
 - [Discovery](#discovery)
 - [Query](#query)
 - [View](#view)
+- [Bulk replace](#bulk-replace)
 - [Testing](#testing)
         - [Docker](#docker)
 - [More Info](#more-info)
@@ -557,6 +558,71 @@ module.exports = function(server) {
     // Alternatively user can also specify the filter for view query
     ds.connector.viewDocs('design_doc', 'view_name', {key: 'filter'}, 
       function(err, results) {});
+  });
+};
+```
+
+# Bulk replace
+
+Given an array of data to be updated, Cloudant supports the idea of performing bulk replace on a model instance.
+
+**Note:** To perform bulk replace, each data in the array data set needs to have the `id` and `_rev` property corresponding to the documents `id` and `_rev` property in the database.
+
+Example:
+
+```server/boot/script.js```
+
+```javascript
+var dataToCreate = [
+  {id: 1, name: 'Foo', age: 1},
+  {id: 2, name: 'Bar', age: 1},
+  {id: 3, name: 'Baz', age: 2},
+  {id: 4, name: 'A', age: 4},
+  {id: 5, name: 'B', age: 5},
+  {id: 6, name: 'C', age: 6},
+  {id: 7, name: 'D', age: 7},
+  {id: 8, name: 'E', age: 8},
+ ];
+var dataToUpdate = [
+ {id: 1, name: 'Foo-change', age: 11},
+ {id: 5, name: 'B-change', age: 51},
+ {id: 8, name: 'E-change', age: 91}
+];
+
+module.exports = function(app) {
+  var db = app.dataSources.cloudantDS;
+  var Employee = app.models.Employee;
+
+  db.once('connected', function() {
+    db.automigrate(function(err) {
+      if (err) throw err;
+
+      Employee.create(dataToCreate, function(err, result) {
+        if (err) throw err;
+        console.log('\nCreated instance: ' + JSON.stringify(result));
+
+        dataToUpdate[0].id = result[0].id;
+        dataToUpdate[0]._rev = result[0]._rev;
+        dataToUpdate[1].id = result[4].id;
+        dataToUpdate[1]._rev = result[4]._rev;
+        dataToUpdate[2].id = result[7].id;
+        dataToUpdate[2]._rev = result[7]._rev;
+
+        // note: it is called `db.connector.bulkReplace`
+        // rather than `Employee.bulkReplace`
+        db.connector.bulkReplace('Employee', dataToUpdate, function(err, result) {
+          if (err) throw err;
+
+          console.log('\nBulk replace performed: ' + JSON.stringify(result));
+
+          Employee.find(function(err, result) {
+            if (err) throw err;
+
+            console.log('\nFound all instances: ' + JSON.stringify(result));
+          });
+        });
+      });
+    });
   });
 };
 ```
--- a/lib/cloudant.js
+++ b/lib/cloudant.js
@@ -551,6 +551,36 @@ Cloudant.prototype.updateAll = function(model, where, data, options, cb) {
 };
 
 /**
+ * Perform a bulk update on a model instance
+ *
+ * @param {String} model The model name
+ * @param {Array} dataList List of data to be updated
+ * @callback {Function} cb The callback function
+ */
+Cloudant.prototype.bulkReplace = function(model, dataList, cb) {
+  debug('Cloudant.prototype.bulkReplace %j %j', model,
+          dataList);
+  var self = this;
+  var mo = self.selectModel(model);
+
+  var dataToBeUpdated = _.map(dataList, function(data) {
+    return self.toDB(model, mo, data);
+  });
+
+  mo.db.bulk({docs: dataToBeUpdated}, function(err, result) {
+    if (err) return cb(err);
+    var errorArray = _.filter(result, 'error');
+    if (errorArray.length > 0) {
+      err = new Error(g.f(util.format('Unable to update 1 or more ' +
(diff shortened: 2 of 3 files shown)


jannyHou:
@ssh24 I think as we discussed, this api expects user to provide the `_rev` right? I assume it would be a very simple wrapper of driver function `db.bulkUpdate()`.

- we don't fetch data and merge `_rev` for them
- what ever user provide in `db.connector.massUpdate` is what we pass in to `db.bulkUpdate()`

README.md
```diff
@@ -561,6 +574,68 @@ module.exports = function(server) {
 };
 ```
 
+# Mass update
+
+Given an array of data to be updated, cloudant supports the idea of performing bulk update on a model instance.
```
dhmlau: `cloudant` --> `Cloudant`

lib/cloudant.js
```diff
@@ -540,11 +540,59 @@ Cloudant.prototype.updateAll = function(model, where, data, options, cb) {
...
-        var res = {};
-        res.count = docs.length;
-        return cb(err, res);
+        var errorArray = _.filter(result, 'error');
+        if (errorArray.length > 0) {
+          err = new Error(g.f(util.format('Unable to update 1 or more ' +
+          'document(s): %s', util.inspect(errorArray, 2))));
+          return cb(err);
+        } else {
+          return cb(err, {count: result.length});
```
dhmlau: do we just want to return the count?

kjdelisle:
@jannyHou makes a good point, we should probably rename the function to `bulkReplace`. It gets rid of the ambiguity, uses a more familiar naming convention and also heavily implies the idea that if you don't provide the **whole** object, you'll lose data.

README.md
```diff
@@ -38,6 +38,7 @@ The `loopback-connector-cloudant` module is the Cloudant connector for the LoopB
 - [Discovery](#discovery)
 - [Query](#query)
 - [View](#view)
+- [Mass update](#mass-update)
```
kjdelisle: Let's call it `Bulk replace` instead.

README.md
```diff
@@ -278,6 +279,18 @@ Let's say we have an instance in the database:
     });
   ```
 
+  - update/updateAll
```
kjdelisle: This should be a different PR; this PR is for `bulkReplace`.

README.md
```diff
@@ -561,6 +574,68 @@ module.exports = function(server) {
 };
 ```
 
+# Mass update
+
+Given an array of data to be updated, cloudant supports the idea of performing bulk update on a model instance.
+
+**Note:** To perform mass update, each data in the array data set needs to have a `_rev` property that matches the corresponding documents `_rev` property in the database.
```
kjdelisle: `mass update` -> `bulk replace`

README.md
```diff
@@ -561,6 +574,68 @@ module.exports = function(server) {
...
+        if (err) throw err;
+        console.log('\nCreated instance: ' + JSON.stringify(result));
+
+        dataToUpdate[0]._rev = result[0]._rev;
+        dataToUpdate[1]._rev = result[4]._rev;
+        dataToUpdate[2]._rev = result[7]._rev;
+
+        // note: it is called `db.connector.massUpdate`
+        // rather than `Employee.massUpdate`
+        db.connector.massUpdate('Employee', dataToUpdate, function(err, result) {
```
kjdelisle: `bulkReplace`

lib/cloudant.js
```diff
@@ -540,11 +540,59 @@ Cloudant.prototype.updateAll = function(model, where, data, options, cb) {
...
+};
+
+/**
+ * Perform a bulk update on a model instance
+ *
+ * @param {String} model The model name
+ * @param {Array} datas List of data to be updated
+ * @callback {Function} cb The callback function
+ */
+Cloudant.prototype.massUpdate = function(model, datas, cb) {
```
kjdelisle: `massUpdate` -> `bulkReplace`

lib/cloudant.js
```diff
@@ -540,11 +540,59 @@ Cloudant.prototype.updateAll = function(model, where, data, options, cb) {
...
+  debug('Cloudant.prototype.massUpdate %j %j',
```
kjdelisle: `bulkReplace`

lib/cloudant.js
```diff
@@ -540,11 +540,59 @@ Cloudant.prototype.updateAll = function(model, where, data, options, cb) {
...
+          datas);
+  var self = this;
+  var mo = self.selectModel(model);
+
+  var toBeUpdatedList = [];
+  var data = {};
+
+  self.all(model, {where: {}}, {raw: true}, function(err, docs) {
```
kjdelisle: Unlike the update/updateAll function, we don't want to call `self.all`, since we're just receiving an array of objects to replace in Cloudant.

ssh24:
> If you don't provide the whole object, you'll lose data.

That is not correct. It will do a partial update on whatever data you send it if the **whole** object is not sent.

PS. I can change the name of the function, that is not a big deal.

lib/cloudant.js
ssh24: Yes, for `updateAll` we just return the count.

lib/cloudant.js
```diff
@@ -540,11 +540,59 @@ Cloudant.prototype.updateAll = function(model, where, data, options, cb) {
...
+    if (err) return cb(err);
+    async.each(docs, function(doc, cb) {
+      data = _.find(datas, {_rev: doc._rev});
+      if (data) {
+        delete doc._rev;
+        _.mergeWith(doc, data, function(dest, src) { return src; });
+        toBeUpdatedList.push(doc);
+      }
```
kjdelisle: If we do this, it means that only items that have matching `_rev` values will be sent to the database.

While this sounds good at first, what it means is that _the user won't know which values were updated_ since they may have had some of their items dropped from the list without warning.

lib/cloudant.js
b-admike: Yeah that's the convention used across the connectors

lib/cloudant.js
```diff
@@ -540,11 +540,59 @@ Cloudant.prototype.updateAll = function(model, where, data, options, cb) {
...
+      });
+    });
+  });
+};
+
+/**
+ * Perform a bulk update on a model instance
+ *
+ * @param {String} model The model name
+ * @param {Array} datas List of data to be updated
```
b-admike: I suggest you use `data` or `dataList` here.

lib/cloudant.js
```diff
@@ -540,11 +540,59 @@ Cloudant.prototype.updateAll = function(model, where, data, options, cb) {
...
+ * @callback {Function} cb The callback function
+ */
+Cloudant.prototype.massUpdate = function(model, datas, cb) {
+  debug('Cloudant.prototype.massUpdate %j %j',
+          datas);
+  var self = this;
+  var mo = self.selectModel(model);
+
+  var toBeUpdatedList = [];
+  var data = {};
```
b-admike: Something along the lines of `dataWithRev` here would be more descriptive

README.md
kjdelisle: Given my other feedback, it's also not the case that `_rev` is required; it's only necessary if you know the record exists and want to replace it.

test/update.test.js
```diff
@@ -157,13 +195,89 @@ describe('updateAll', function() {
...
+      testUtil.hasResult(err, result).should.be.ok();
+      dataToBeUpdated[0]._rev = result[0]._rev;
+      dataToBeUpdated[1]._rev = result[3]._rev;
+      dataToBeUpdated[2]._rev = result[5]._rev;
+      db.connector.massUpdate('Product', dataToBeUpdated,
+      function(err, result) {
+        err = testUtil.refinedError(err, result);
+        if (err) return done(err);
+        testUtil.hasResult(err, result).should.be.ok();
+        should.equal(result.length, dataToBeUpdated.length);
```
b-admike: Do a find here to check everything is good?

ssh24: Good catch. Added a `find` call underneath.

lib/cloudant.js
```diff
@@ -551,6 +551,36 @@ Cloudant.prototype.updateAll = function(model, where, data, options, cb) {
...
 
 /**
+ * Perform a bulk update on a model instance
+ *
+ * @param {String} model The model name
+ * @param {Array} dataList List of data to be updated
+ * @callback {Function} cb The callback function
+ */
+Cloudant.prototype.bulkReplace = function(model, dataList, cb) {
+  debug('Cloudant.prototype.massUpdate %j %j', model,
```
kjdelisle: change message to `bulkReplace`

PR #151, opened 2017-06-07
Fix update/updateAll functions

ssh24:
### Description

Return the result in `updateAll` in a correct manner.  Check for error on update if any exists.

connect to #35

### DIFF
--- a/README.md
+++ b/README.md
@@ -279,6 +279,29 @@ Let's say we have an instance in the database:
     });
   ```
 
+  - update/updateAll
+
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -278,6 +278,18 @@ Let's say we have an instance in the database:
     });
   ```
 
+  - update/updateAll
+
+  ```javascript
```
b-admike: You could also show an example without _rev here or note that it could be done

PR #153, opened 2017-06-08
Couchdb 

jannyHou:
## PoC Overview

This is a PoC PR that proves the basic possibility of supporting both couchdb&clouant connector in loopback:

- connect to couchdb database through nano driver
- automigrate a loopback model
- insert document in to database
  - it implies CRUD operations are also possible
- query model instance
  - it implies relation traverse and scope is also possible
  - in couchdb/cloudant, being able to find everything implies you can also delete everything 
  - nested property's query: the test doesn't cover it but I tried with couchdb dashboard, as long as user properly creates index for it, e.g. `['address.tag']`, they can query it like `{address.tag: {$elemMatch: {name: 'tag1'}}}`, same as Cloudant
- delete index
  - it implies we can clean up existing model

## Design (please ready design first then review code and run test, thanks :)
https://github.com/strongloop/loopback-connector-cloudant/pull/153#issuecomment-307177910

## Run Test

- provide `url` in https://github.com/strongloop/loopback-connector-cloudant/pull/153/files#diff-ad1e400f926c720a1d0e73501896e26cR17
- `DEBUG=loopback:connector:couchdb mocha test/couchdb/couchdb-basic.test.js` run tests
- two test suites in `couchdb-basic.test.js`, run them separately(DON'T run both at same time, to save time I didn't write the logic to clean up the existing model&data)

## FOR DEMO

- ONLY 3 files are related. others file changes are caused by file location change.
- lots of things not implemented, just for demo
- temporary naming convention, both **model index** and **property index** should be **configurable**.
- haven't implemented model CRUD functions, just have
  - couchdb insert
  - couchdb index create
  - couchdb index delete


jannyHou:
## Design

### Model
  - create index 
    ```json
    {
      "ddoc": "loobpack__model__name__ddoc", 
      "name": "loopback__model__name",
      "index": {
         "fields": ["loopback__model__name"]
       }
       "type": "json"
    }
    ```

rmg:
I haven't looked at the code, but based on the design, here are some things to keep in mind:
 - long document and property names result in larger indexes, which mean more I/O, which means slower performance
 - all index/map/reduce updates do design-document level locking, so it is often better to create multiple design documents instead of putting multiple indexes under a single doc
 - the example uses `loopback__model__name`, but existing db's are likely to already be using `type`, `model`, or `class` for this property, so it should probably be configurable

jannyHou:
@rmg thank you so much!

> long document and property names result in larger indexes, which mean more I/O, which means slower performance
AND
   the example uses loopback__model__name, but existing db's are likely to already be using type, model, or class for this property, so it should probably be configurable

Yes both model index and property index should be configurable.  Good to learn that "long document and property names result in larger indexes" :+1: 


> all index/map/reduce updates do design-document level locking, so it is often better to create multiple design documents instead of putting multiple indexes under a single doc

This is a good news for creating separate design document for model properties XD

jannyHou:
An important point from @rmg that might affect the design:

With 0 index design doc, couchdb provides an "all_fields" indexable query system, which saves all index creating/deleting effort.

This requires that the database could not have any index design doc. Considering the current design requires no multiple fields index overriding a single field index, it requires same level "careful" for users.

jannyHou:
A good news:
It proves the limitations I run into are caused by an old version of Couchdb I used, after upgrading to the latest Couchdb, the index system does work as how Cloudant works.

> _find chooses which index to use for responding to a query, unless you specify an index at query time.
If there are two or more json type indexes on the same fields, the index with the smallest number of fields in the index is preferred. If there are still two or more candidate indexes, the index with the first alphabetical name is chosen.

⬆️  explains everything and I verified it.

You can specify an index by `{'use_index': [<designdoc_name>]}`, if you don't, **ALL** fields are indexable no matter what custom indexes user creates.

I will add more tests to prove it as a PoC should do.

It also implies, we can use most of existing code in Cloudant connector by removing two phases:
  - create index design doc for models
  - specify model index when perform query
It is even simpler than Cloudant.

jannyHou:
Create the couchdb connector in https://github.com/strongloop/loopback-connector-cloudant/pull/155/commits/0b3b94bb563f1a8efb56c5f0e50c759a424ecd08


## Main Block
- sort still requires creating index for field

> To use sorting, ensure that:
    At least one of the sort fields is included in the selector.
    There is an index already defined, with all the sort fields in the same order.
    Each object in the sort array has a single key.

jannyHou:
Closing it since PoC is done, we are building the connector now.

PR #154, opened 2017-06-09
Do not strip rev value on create

ssh24:
### Description
Currently, we as the ORM strips off the `_rev` value before doing a create operation. This is not needed since, cloudant ignores the `_rev` value when passed into the payload data for its own clarity. So if the user specifies the `_rev` value, **it is upto them to know that value won't be accepted as cloudant will automatically generate a unique _rev value**

connect to https://github.com/strongloop-internal/scrum-apex/issues/234

### DIFF
--- a/test/create.test.js
+++ b/test/create.test.js
@@ -37,6 +37,27 @@ describe('create', function() {
       db.automigrate(done);
     });
   });
+
+  it('creates a model instance when `_rev` is provided', function(done) {
+    var newBread = _.cloneDeep(bread);
+    newBread._rev = '1-somerandomrev';
+    Product.create(newBread, function(err, result) {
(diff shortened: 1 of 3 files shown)


test/create.test.js
```diff
@@ -37,6 +37,28 @@ describe('create', function() {
...
     });
   });
+
+  it('creates a model instance when `_rev` is provided', function(done) {
+    var newBread = _.cloneDeep(bread);
+    newBread._rev = '1-somerandomrev';
+    Product.create(newBread, function(err, result) {
+      err = testUtil.refinedError(err, result);
+      if (err) return done(err);
+      console.log('Result: ' + JSON.stringify(result));
```
b-admike: Remove console.log and check for result here.

PR #156, opened 2017-06-15
Dev/couchdb

jannyHou:
### Description

Branch `dev/couchdb` is to guarantee in the short term:
  - The dev of couchdb won't block cloudant improvement, and couchdb gets sync with cloudant patches/features. 
  - New PRs for couchdb/cloudant are tested. 

### Tips

- On branch `dev/couchdb`, I moved most of current `cloudant.js` code into `couchdb.js`, cloudant will extend that file after the refactor.
- I created a new branch in juggler for couchdb tests, a few tests fail and I need to investigate, and that branch only tests 3 files atm: basic-query, relation, include. CRUD tests need more time due to `_rev`, other than CRUD I think we cover most important ones.
- This is just a short term solution, next we will `have couchdb repo`  ---> `cloduant extends couchdb, both code and tests` ---> `continue recovering juggler tests`

### Review

- The first commit is just created for compare(easier review)

- The second commit [couchdb connector](https://github.com/strongloop/loopback-connector-cloudant/pull/156/commits/8e1c8705e9e1a99c1dfa93a235e2b28f0fc2796a) shows the difference between cloudant and couchdb
  - Majority of code in function `connect` and `constructor` get removed in this commit but they are brought back in commit [Fix url test](https://github.com/strongloop/loopback-connector-cloudant/pull/156/commits/3338df7d45ae036d57cdca3c0d682c435d7f4fdf), don't be terrified :=p

- It tests with a juggler branch `couchdb/test`, the reason is I could not fix all juggler test&refactor the juggler tests in a few days, but I still fixed something and want to have new patches/features get tested with juggler. Eventually `couchdb/test` should be replaced by `master`

- Have another story to track those few failures in juggler:
https://github.com/strongloop/loopback-connector-cloudant/issues/33#issuecomment-308217081

### DIFF
--- a/test.js
+++ b/test.js
@@ -27,7 +27,7 @@ var CONNECT_DELAY = ms('5s');
 var containerToDelete = null;
 
 async.waterfall([
-  dockerStart('klaemo/couchdb:2.0-dev'),
+  dockerStart('klaemo/couchdb:2.0.0'),
   sleep(ms('2s')),
   setCloudantEnv,
   waitFor('/_all_dbs'),
@@ -69,7 +69,13 @@ function dockerStart(imgName) {
         console.log('starting container from image: %s', imgName);
         docker.createContainer({
           Image: imgName,
-          Cmd: '-d -p 5984:5984 klaemo/couchdb:2.0-dev --with-haproxy --admin=admin:pass',
+          HostConfig: {
+            PublishAllPorts: true,
+          },
+          Env: [
+            'COUCHDB_USER=' + process.env.COUCHDB_USERNAME,
+            'COUCHDB_PASSWORD=' + process.env.COUCHDB_PASSWORD,
+          ],
         }, function(err, container) {
           console.log('recording container for later cleanup: ', container.id);
           containerToDelete = container;
@@ -90,11 +96,10 @@ function setCloudantEnv(container, next) {
     // if swarm, Node.Ip will be set to actual node's IP
     // if not swarm, but remote docker, use docker host's IP
     // if local docker, use localhost
-    console.log(c);
     var host = _.get(c, 'Node.IP', _.get(docker, 'modem.host', '127.0.0.1'));
     // container's port 80 is dynamically mapped to an external port
     var port = _.get(c,
-      ['NetworkSettings', 'Ports', '80/tcp', '0', 'HostPort']);
+      ['NetworkSettings', 'Ports', '5984/tcp', '0', 'HostPort']);
     process.env.COUCHDB_PORT = port;
     process.env.COUCHDB_HOST = host;
     process.env.COUCHDB_URL = 'http://' + host + ':' + port;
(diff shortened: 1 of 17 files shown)


test.js
```diff
@@ -69,9 +69,7 @@ function dockerStart(imgName) {
         console.log('starting container from image: %s', imgName);
         docker.createContainer({
           Image: imgName,
-          HostConfig: {
-            PublishAllPorts: true,
-          },
+          Cmd: '-d -p 5984:5984 klaemo/couchdb:2.0-dev --with-haproxy --admin=admin:pass',
```
rmg: Don't add `Cmd`, instead add this:
```js
Env: [
  'COUCHDB_USER=' + process.env.COUCHDB_USERNAME,
  'COUCHDB_PASSWORD=' + process.env.COUCHDB_PASSWORD,
],
```

test.js
```diff
@@ -92,20 +90,21 @@ function setCloudantEnv(container, next) {
     // if swarm, Node.Ip will be set to actual node's IP
     // if not swarm, but remote docker, use docker host's IP
     // if local docker, use localhost
+    console.log(c);
     var host = _.get(c, 'Node.IP', _.get(docker, 'modem.host', '127.0.0.1'));
     // container's port 80 is dynamically mapped to an external port
     var port = _.get(c,
       ['NetworkSettings', 'Ports', '80/tcp', '0', 'HostPort']);
```
rmg: Change `'80/tcp'` to `'4984/tcp'`.

test.js
```diff
@@ -11,23 +11,23 @@ var ms = require('ms');
...
+process.env.COUCHDB_URL = 'TBD';
 
 var CONNECT_RETRIES = 30;
 var CONNECT_DELAY = ms('5s');
 
 var containerToDelete = null;
 
 async.waterfall([
-  dockerStart('ibmcom/cloudant-developer:2.0.1'),
+  dockerStart('klaemo/couchdb:2.0-dev'),
```
rmg: Should be using `klaemo/couchdb:2.0.0` instead.

test.js
```diff
@@ -69,9 +69,7 @@ function dockerStart(imgName) {
         console.log('starting container from image: %s', imgName);
         docker.createContainer({
           Image: imgName,
-          HostConfig: {
```
rmg: Leave this here.

test.js
rmg: For reference, I pulled these env vars directly from the startup script since they aren't described very clearly in the actual `README.md`: https://github.com/apache/couchdb-docker/blob/master/2.0.0/docker-entrypoint.sh

jannyHou:
@slnode test please

jannyHou:
Last week the tests fail when creating index(unauthorized error), but when I pull the new image again tests pass, hope it could solve the problem.

jannyHou:
good catch.

PR #159, opened 2017-07-14
Recover manipulation.test.js

jannyHou:
connect to  strongloop/loopback-connector-cloudant#33

related to https://github.com/strongloop/loopback-datasource-juggler/pull/1420

### DIFF
--- a/lib/cloudant.js
+++ b/lib/cloudant.js
@@ -173,6 +173,13 @@ Cloudant.prototype._insert = function(model, data, cb) {
       return cb(err);
     }
     data[idName] = result.id;
+
+    // Convert ID to Number if Model defines ID as type Number
+    var idType = mo.mo.properties[idName].type.name;
+    if (idType === 'Number') {
+      result.id = parseInt(result.id);
+    }
+
     cb(null, result.id, result.rev);
   });
 };
@@ -359,7 +366,9 @@ Cloudant.prototype.buildSort = function(mo, model, order) {
 Cloudant.prototype.destroy = function destroy(model, id, options, cb) {
   var mo = this.selectModel(model);
   this.all(model, {where: {id: id}}, {raw: true}, function(err, doc) {
-    if (doc.length > 1) cb(new Error(
+    if (err) return cb(err);
+
+    if (doc.length > 1) return cb(new Error(
       'instance method destroy tries to delete more than one item!'));
     else if (doc.length === 1) {
       mo.db.destroy(doc[0]._id, doc[0]._rev, function(err, result) {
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -33,7 +33,7 @@
     "dockerode": "^2.4.3",
     "eslint": "^2.13.1",
     "eslint-config-loopback": "^4.0.0",
-    "loopback-datasource-juggler": "^3.0.0",
+    "loopback-datasource-juggler": "git://github.com/strongloop/loopback-datasource-juggler.git#cloudant/crud-test",
```
jannyHou: Make sure I don't forget to change it back

lib/cloudant.js
```diff
@@ -173,6 +173,13 @@ Cloudant.prototype._insert = function(model, data, cb) {
       return cb(err);
     }
     data[idName] = result.id;
+
+    // Convert ID to Number if Model defines ID as type Number
```
jannyHou: Test cases: https://github.com/strongloop/loopback-datasource-juggler/blob/master/test/manipulation.test.js#L2297-L2384
Those tests are using `Number` type Id

lib/cloudant.js
```diff
@@ -368,7 +377,7 @@ Cloudant.prototype.destroy = function destroy(model, id, options, cb) {
         cb(err, result && result.ok ? {count: 1} : {count: 0});
       });
     } else {
-      cb(new Error('could not find matching item in database!'));
+      return cb(null, {count: 0});
```
jannyHou: Loopback connectors doesn't return error if item not found in database. It just return the `count` and juggler return error according to the flag `strictDelete` and the `count`, see:
https://github.com/strongloop/loopback-datasource-juggler/blob/3a10c77cbdeb658ffc34e8db95acc7c57f2010c0/lib/dao.js#L2819

Related test cases: https://github.com/strongloop/loopback-datasource-juggler/blob/3a10c77cbdeb658ffc34e8db95acc7c57f2010c0/test/manipulation.test.js#L1837-L1880

test/imported.test.js
```diff
@@ -17,4 +17,5 @@ describe('cloudant imported features', function() {
   require('loopback-datasource-juggler/test/hooks.test.js');
   require('loopback-datasource-juggler/test/relations.test.js');
   require('loopback-datasource-juggler/test/include.test.js');
+  require('loopback-datasource-juggler/test/manipulation.test.js');
```
jannyHou: The above is same as require `common.batch.test.js` and `include.test.js`(couchdb2 uses those two instead of listing them all), just since it's extending everything from couchdb2 soon....

ssh24:
@slnode test please

PR #163, opened 2017-07-21
Inherit couchdb functionalities

ssh24:
### Description

Remove duplicate functions and allow to be inherited from couchdb. Leave the functions that are needed specifically on Cloudant.

connect to strongloop-internal/scrum-apex#258

**Note:** CI is failing because the CouchDB module is not released yet. Hence, the PR *cannot be* technically merged till CouchDB is released.


package.json
```diff
@@ -26,6 +26,7 @@
     "debug": "^2.2.0",
     "lodash": "^4.17.4",
     "loopback-connector": "^4.0.0",
+    "loopback-connector-couchdb2": "https://github.com/strongloop/loopback-connector-couchdb2.git",
```
kjdelisle: I've released the beta for the CouchDB connector @ 0.9.0, so please change this accordingly!

ssh24:
@slnode test please

jannyHou:
`Cloudant.prototype,all` injects `use_index` field into the query, while after the index PR from couchdb checked in, that line needs to be removed.

lib/cloudant.js
```diff
@@ -6,7 +6,7 @@
 'use strict';
 
 var g = require('strong-globalize')();
-var Connector = require('loopback-connector').Connector;
+var CouchDB = require('loopback-connector-couchdb2').CouchDB;
```
jannyHou: We need them in Cloudant too.
```
const DEFAULT_MODEL_VIEW = 'loopback__model__name';
const DEFAULT_MODEL_PREFIX = 'LBModel';
const DEFAULT_PROPERTY_PREFIX = 'LBIndex';
```

jannyHou:
We need to inherit the `sort` function as well now... sorting syntax differs:
- text index: `{sort: [{'name:string': 'asc'}]}`
- json index: `{sort: [{'name': 'asc'}]}`

So if user want to query with a text index:
- create that index
- specify `use_index` in query
- specify field's type in query.order

lib/cloudant.js
ssh24: Do we? I thought the only functions needing them is all in Couch so it will use the one in there.

package.json
```diff
@@ -18,15 +18,15 @@
...
+    "mocha": "./node_modules/.bin/_mocha --timeout 40000 --require strong-mocha-interfaces --require test/init.js --ui strong-bdd"
   },
   "dependencies": {
     "async": "^1.5.0",
     "cloudant": "^1.4.0",
     "debug": "^2.2.0",
     "lodash": "^4.17.4",
     "loopback-connector": "^4.0.0",
-    "loopback-connector-couchdb2": "^1.0.0",
+    "loopback-connector-couchdb2": "https://github.com/strongloop/loopback-connector-couchdb2.git",
```
ssh24: Need to do this for now. Need another `patch` version release for the couchdb.

test/cloudant.test.js
```diff
@@ -372,9 +389,13 @@ describe('cloudant connector', function() {
             should.not.exist(err);
             should.exist(result);
             should.equal(result.length, 3);
-            testUtil.checkData(data[0], result[0].__data);
-            testUtil.checkData(data[1], result[1].__data);
-            testUtil.checkData(data[2], result[2].__data);
+            // checkData ignoring its order
+            data.forEach(function(item, index) {
+              var r = _.find(result, function(o) {
+                return o.__data.id === item.id;
```
b-admike: I think `o.toObject().id === item.id` might also work here

test/cloudant.test.js
```diff
@@ -372,9 +389,13 @@ describe('cloudant connector', function() {
...
+              });
+              testUtil.checkData(data[index], r.__data);
```
b-admike: `r.toObject()` can also work here

test/cloudant.test.js
```diff
@@ -255,12 +268,16 @@ describe('cloudant connector', function() {
...
               if (err) return done(err);
               customers.length.should.be.equal(2);
-              customers[0].address.city.should.be.eql('San Mateo');
-              customers[1].address.city.should.be.eql('San Jose');
+              var expected1 = ['San Mateo', 'San Jose'];
+              var expected2 = ['San Jose', 'San Mateo'];
+              var actual = customers.map(function(c) {
+                return c.address.city;
+              });
+              should(actual).be.oneOf(expected1, expected2);
```
jannyHou: @ssh24 sorry I am confused here, we specify the order as `address.city DESC` in query, and this test aims to make sure the result returned in the right sequence right?

But `should(actual).be.oneOf(expected1, expected2);` means you check **regardless** the order

kjdelisle: Excellent point; the test should ascertain order of items.

test/cloudant.test.js
```diff
@@ -230,19 +243,19 @@ describe('cloudant connector', function() {
...
+        order: 'address.state DESC'},
           function(err, customers) {
             should.exist(err);
-            err.message.should.match(/Unspecified or ambiguous sort type/);
+            err.message.should.match(/no_usable_index,missing_sort_index/);
             done();
           });
       });
-      it('returns result when sorting type provided - missing first level' +
+      it('returns result when sorting type provided - missing first level ' +
```
jannyHou: see comment https://github.com/strongloop/loopback-connector-cloudant/pull/163/files#r134341551
I think you might be affected by the code in this test case, while this one aimed to make sure my "field type detecting" algorithm in `buildSort` doesn't throw error if the property doesn't exist in modelDef :-p

jannyHou:
@ssh24 I've seen the tests pass on your local while still fails on CI...could you double check?

package.json
```diff
@@ -18,15 +18,15 @@
...
+    "mocha": "./node_modules/.bin/_mocha --timeout 40000 --require strong-mocha-interfaces --require test/init.js --ui strong-bdd"
   },
   "dependencies": {
     "async": "^1.5.0",
     "cloudant": "^1.4.0",
     "debug": "^2.2.0",
     "lodash": "^4.17.4",
     "loopback-connector": "^4.0.0",
-    "loopback-connector-couchdb2": "^1.0.0",
+    "loopback-connector-couchdb2": "https://github.com/strongloop/loopback-connector-couchdb2.git#recover-skipped-test",
```
ssh24: This needs to be pointed to the latest patch once the [Couch PR](https://github.com/strongloop/loopback-connector-couchdb2/pull/31) lands.

lib/cloudant.js
jannyHou: ah, you are right!

lib/cloudant.js
```diff
@@ -99,1155 +99,88 @@ Cloudant.prototype.connect = function(cb) {
...
- */
-function isDate(type) {
-  return (type === Date || type === 'date' || type === 'Date');
-}
+Cloudant.prototype.getModelObjectSettings = function(mo) {
+  if (mo) return mo.settings.cloudant;
+  return undefined;
+};
+
+// /**
```
jannyHou: @ssh24 last comment: just remove it, we don't need this function anymore.

Issue #165, opened 2017-07-25
Use built in _id field and index


ritch:
Looks like we are not making it possible to use the built in `_id` field and index and instead create a separate index and alias. We should support disabling this separate index.

kjdelisle:
This issue is being raised as a performance issue, but whoever ends up taking this task should run a performance test with two different versions of the same model:
- One where the id is autogenerated (`id`)
- One where you explicitly declare `_id` as a field, and mark it with `{ id: true }`

If this doesn't yield a statistically significant difference in performance, then we'll need to re-evaluate the importance of this issue.

kjdelisle:
So, I've run some automated tests with both these scenarios against the LoopBack Notes template, on Cloudant Local and found no statistically significant difference in performance between them. I'm not sure where the performance complaint stems from in that respect.

I think as a workaround, users can explicitly define `_id` as a field on their models for the loopback-connector-cloudant in the 1.x release. There's a bug with this version that currently removes the `_id` field if present, which will break this approach, but we have a PR to address this: https://github.com/strongloop/loopback-connector-cloudant/pull/170

kjdelisle:
The greater problem stems from the idea that LoopBack 3 itself is the reason for the default `id` alias, which is not something that can be controlled for at this level.

kjdelisle:
Fixed and released!

PR #168, opened 2017-07-26
Fix to return _id to when it is model id

virkt25:
Fixes an issue where if _id was set as id on model, any query won’t
return _id anymore.

### DIFF
--- a/test/id.test.js
+++ b/test/id.test.js
@@ -1,0 +1,52 @@
+// Copyright IBM Corp. 2016. All Rights Reserved.
+// Node module: loopback-connector-cloudant
+// This file is licensed under the Artistic License 2.0.
+// License text available at https://opensource.org/licenses/Artistic-2.0
+
+'use strict';
+
+var Cloudant = require('../lib/cloudant');
+var should = require('should');
+var db, Task;
+
+describe('using _id as id', function() {
+  before(function(done) {
+    db = getDataSource();
+
+    Task = db.define('Task', {
+      task: {type: String},
+      owner: {type: String},
+      _id: {type: String, id: true},
+    }, {forceId: true});
+
+    db.automigrate('Task', done);
+  });
+
+  after(function(done) {
+    Task.destroyAll(done);
+  });
+
+  it('result should include _id', function(done) {
+    var ownerName = 'John Smith';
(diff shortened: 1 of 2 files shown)


test/id.test.js
```diff
@@ -0,0 +1,52 @@
...
+      _id: {type: String, id: true},
+    }, {forceId: true});
+
+    Task.destroyAll(function(err) {
+      done();
+    });
+  });
+
+  after(function(done) {
+    Task.destroyAll(function() {
```
ssh24: `Task.destroyAll(done)`

test/id.test.js
```diff
@@ -0,0 +1,52 @@
...
+  before(function(done) {
+    db = getDataSource();
+
+    Task = db.define('Task', {
+      task: {type: String},
+      owner: {type: String},
+      _id: {type: String, id: true},
+    }, {forceId: true});
+
+    Task.destroyAll(function(err) {
```
ssh24: Why are you doing a destroyAll here?

b-admike: Do we need this call for `destroyAll()` here?

virkt25:
connect to #165

test/id.test.js
virkt25: Changed to automigrate.

virkt25: Changed to automigrate.

test/id.test.js
```diff
@@ -0,0 +1,48 @@
...
+  });
+
+  it('result should include _id', function(done) {
+    var ownerName = 'John Smith';
+    Task.create({
+      task: 'some random task',
+      owner: ownerName,
+    }, function(err, task) {
+      if (err) return done(err);
+      var taskId = task._id;
```
ssh24: I would assert for `task` and `task._id` first.
```js
should.exist(task);
should.exist(task._id);
```

kjdelisle:
+ 1 to @ssh24's remaining suggestion; after that, LGTM.

kjdelisle:
Node 0.12 and 0.12 support needs further discussion, but it's not a real blocker for this fix.

kjdelisle:
All relevant tests were passing, ignoring the above.

PR #174, opened 2017-08-23
Add advisory docs on forceId

ssh24:
### Description

If there are two models using `_id` as pk, and with `forceId: false`, there is a possibility of a document conflict.

For example:
```
{_id: 'myid', loopback__model__name: 'Foo'}
{_id: 'myid', loopback__model__name: 'Bar'}
```

This PR adds advisory docs on the behaviour of using `forceId` and leaves the onus on the user to use the flag.

fixes https://github.com/strongloop-internal/scrum-apex/issues/255

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@ The `loopback-connector-cloudant` module is the Cloudant connector for the LoopB
         - [Model-specific Configuration](#model-specific-configuration)
         - [_rev Property](#_rev-property)
             - [Example CRUD operations with `_rev`](#example-crud-operations-with-_rev)
+        - [Force Id](#force-id)
 - [Setup Cloudant Instance](#setup-cloudant-instance)
 - [Installation](#installation)
 - [Configuration](#configuration)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -545,6 +546,141 @@ module.export = function migrateData(app) {
 }
 ```
 
+# Force Id
+
+**Note:** If you have two models that have a customized `_id` as primary key set with ```"forceId": false``` in the model definition, it will result in a conflict.
```
b-admike: Suggestion: `two models that have the same custom '_id'...`

README.md
```diff
@@ -545,6 +546,141 @@ module.export = function migrateData(app) {
 }
 ```
 
+# Force Id
+
```
b-admike: Maybe we can give them a brief description of what `forceId` flag does or link them to our doc if it already exists.

README.md
```diff
@@ -545,6 +546,141 @@ module.export = function migrateData(app) {
...
+```
+{_id: 'myid', loopback__model__name: 'Foo'}
+{_id: 'myid', loopback__model__name: 'Bar'}
+```
+
+It is recommended to not set ```"forceId": false``` because data belonging to different models can interfere with each other.
+
+Here is an example of a case that would result in a document update conflict using ```"forceId": false```:
+
+### model-definition
```
b-admike: `model-definitions`

README.md
```diff
@@ -545,6 +546,141 @@ module.export = function migrateData(app) {
...
+Automigrate completed
+Created employee instance: [ { _id: '1', name: 'Foo' }, { _id: '2', name: 'Bar' } ]
+
+/Users/ssh/workspace/sandbox/loopback-sandbox/apps/cloudant/cloudant-forceId-app/server/boot/script.js:33
+          if (err) throw err;
+                   ^
+Error: Document update conflict. (duplicate?),Error: Document update conflict. (duplicate?)
+```
+
+
```
b-admike: To wrap up, you can mention how to correct the above example here. They need to at least set`forceId: true` on one of the model defs or take our recommendation and set it to `true` on both (or dont set it if that is the default behaviour).

README.md
```diff
@@ -545,6 +546,141 @@ module.export = function migrateData(app) {
...
+Web server listening at: http://localhost:3000
+Browse your REST API at http://localhost:3000/explorer
+
+Automigrate completed
+Created employee instance: [ { _id: '1', name: 'Foo' }, { _id: '2', name: 'Bar' } ]
+
+/Users/ssh/workspace/sandbox/loopback-sandbox/apps/cloudant/cloudant-forceId-app/server/boot/script.js:33
+          if (err) throw err;
+                   ^
+Error: Document update conflict. (duplicate?),Error: Document update conflict. (duplicate?)
```
b-admike: I wonder why we get the same error message twice here

ssh24: It prints out twice because there are two instances being created and thrown the error :P

README.md
ssh24: Done. See the next line.

README.md
```diff
@@ -545,6 +546,147 @@ module.export = function migrateData(app) {
 }
 ```
 
+# Force Id
+
+Learn more about LoopBack's [forceId](https://loopback.io/doc/en/lb3/Model-definition-JSON-file.html).
+
+Every document stored in Cloudant has a unique id, stored as `_id`, and the database has built-in index for it. Retrieving data by its `_id` gives a better performance than querying by other field.
+
+If you have two models that allow customized `_id` as primary key set with ```"forceId": false``` in the model definition, it will result in a document update conflict.
```
b-admike: `it could result in a document update conflict.`

README.md
```diff
@@ -545,6 +546,147 @@ module.export = function migrateData(app) {
 }
 ```
 
+# Force Id
+
+Learn more about LoopBack's [forceId](https://loopback.io/doc/en/lb3/Model-definition-JSON-file.html).
+
+Every document stored in Cloudant has a unique id, stored as `_id`, and the database has built-in index for it. Retrieving data by its `_id` gives a better performance than querying by other field.
+
```
jannyHou: Add one more line:
Therefore, the connector always assign the `_id` field as loopback model's primary key.

README.md
jannyHou: We recommend user to be careful when creating customized model `id` instead of using the auto-generated one.

README.md
```diff
@@ -545,6 +546,147 @@ module.export = function migrateData(app) {
 }
 ```
 
+# Force Id
```
jannyHou: Can we move this part under the model? It's more about the model design:

- Model
         - Map Between Model And Document
         - Map Between Model And Document
         - Model-specific Configuration		        
         - _rev Property
        ... ...
         - Force Id

README.md
```diff
@@ -545,6 +546,147 @@ module.export = function migrateData(app) {
...
+
+Learn more about LoopBack's [forceId](https://loopback.io/doc/en/lb3/Model-definition-JSON-file.html).
```
jannyHou: Maybe add one sentence: 
Briefly, in loopback `forceId` means user can specify the value of the primary key when creating a model instance, instead of using an auto-generated one.

README.md
```diff
@@ -302,6 +303,146 @@ Let's say we have an instance in the database:
...
+
+If you have two models that allow customized `_id` as primary key set with ```"forceId": false``` in the model definition, it could result in a document update conflict.
+
+For example:
+```
+{_id: 'myid', loopback__model__name: 'Foo'}
+{_id: 'myid', loopback__model__name: 'Bar'}
+```
+
+We recommend user to be careful when creating customized model id instead of using the auto-generated one because data belonging to different models can interfere with each other.
```
jannyHou: wait...I am not sure why git shows I added on this line, it should be after line 308:
https://github.com/strongloop/loopback-connector-cloudant/pull/174/files#diff-04c6e90faac2675aa89e2176d2eec7d8R308, the sequence would be like this:

In loopback `forceId` means user can specify the value of the primary key when creating a model instance, instead of using an auto-generated one. Learn more about LoopBack's [forceId](https://loopback.io/doc/en/lb3/Model-definition-JSON-file.html).

We recommend user to be careful when creating customized model id instead of using the auto-generated one because data belonging to different models can interfere with each other.

ssh24:
My initial instinct was to create a similar PR for CouchDB. However, after talking to @jannyHou we will create a story to refactor the two repo's as there is a lot of duplicates.

Issue #175, opened 2017-08-24
Update of the connector to Version 2.0.2 breaks connection


sebastiansutter:
I had a working demo in APIC with the loopback connector just before version 2.0.2 got released. When I updated the connector from version 1.2.5 to version 2.0.2, the connection from the data source in APIC Designer is no longer possible.

Steps to reproduce:
create a new loopback project on commandline: apic loopback
open api designer: apic edit
Got to Data Models and create a Cloudant Data Source. Press the link to install the connector

Paste a connection URL and hit save

# Expected result
Data source connection test succeeded

# Actual result
Data Source Connection Failed 
undefined
Data source connection test failed with error: Cannot read property 'db' of undefined

jannyHou:
@sebastiansutter I think I know why it fails, cloudant@1.x sometimes runs into a bug that executes functions before connection established and a valid driver instance get returned. To fix it, in cloudant@2.x we listen on the `connected` event when establishing a connection, for example, if your cloudant@1.x code is written in this way:
```js
cloudantDS.ping(function(err) {
   // Do stuff
});
```
In cloudant@2.x you need to rewrite the function to: 
```js
cloudantDS.once('connected', function() {
  cloudantDS.ping(function(err) {
     // Do stuff
  });
});
```

API designer calls the `ping` function without waiting until `connected`, which means this needs to happen inside connector's `ping` function.
We will fix it asap whenever we have time.

jannyHou:
should be fixed in https://github.com/strongloop/loopback-connector-cloudant/commit/b59747cc36709d97087ded502f63ed3044b7856b

dhmlau:
Version 2.0.3. has been released.  Closing this issue.

PR #177, opened 2017-09-11
Fix/ping

jannyHou:
connect to strongloop/loopback-connector-cloudant#175
Test case modified in PR https://github.com/strongloop/loopback-connector-couchdb2/pull/33

### DIFF
--- a/lib/cloudant.js
+++ b/lib/cloudant.js
@@ -124,7 +124,7 @@ Cloudant.prototype.ping = function(cb) {
     });
   } else {
     self.dataSource.once('connected', function(err) {
-      if (err) cb('ping failed');
+      if (err) cb(new Error('ping failed'));
       self.getDriverInst().db.list(function(err, result) {
         returnCB(err, result);
       });
@@ -133,7 +133,7 @@ Cloudant.prototype.ping = function(cb) {
 
   function returnCB(err, result) {
     debug('Cloudant.prototype.ping results %j %j', err, result);
-    if (err) cb('ping failed');
+    if (err) cb(new Error('ping failed'));
     else cb();
   }
 };
--- a/lib/cloudant.js
+++ b/lib/cloudant.js
@@ -114,5 +114,31 @@ Cloudant.prototype.getModelObjectSettings = function(mo) {
   return undefined;
 };
 
+Cloudant.prototype.ping = function(cb) {
+  debug('Cloudant.prototype.ping');
+  var self = this;
+  var driverInst = self.getDriverInst();
+  if (driverInst) {
+    driverInst.db.list(function(err, result) {
+      debug('Cloudant.prototype.ping results %j %j', err, result);
+      if (err) cb('ping failed');
+      else cb();
+    });
+  } else {
+    self.once('connected', function(err) {
+      if (err) cb('ping failed');
+      self.getDriverInst().db.list(function(err, result) {
+        debug('Cloudant.prototype.ping results %j %j', err, result);
+        if (err) cb('ping failed');
+        else cb();
+      });
+    });
+  }
+  // this.getDriverInst().db.list(function(err, result) {
+  //   debug('CouchDB.prototype.ping results %j %j', err, result);
+  //   if (err) cb('ping failed');
+  //   else cb();
+  // });
+}
 // mixins
 require('./view')(Cloudant);
--- a/lib/cloudant.js
+++ b/lib/cloudant.js
@@ -120,25 +120,22 @@ Cloudant.prototype.ping = function(cb) {
   var driverInst = self.getDriverInst();
   if (driverInst) {
     driverInst.db.list(function(err, result) {
-      debug('Cloudant.prototype.ping results %j %j', err, result);
-      if (err) cb('ping failed');
-      else cb();
+     returnCB(err, result);
     });
   } else {
     self.dataSource.once('connected', function(err) {
       if (err) cb('ping failed');
       self.getDriverInst().db.list(function(err, result) {
-        debug('Cloudant.prototype.ping results %j %j', err, result);
-        if (err) cb('ping failed');
-        else cb();
+        returnCB(err, result);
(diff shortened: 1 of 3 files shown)


lib/cloudant.js
```diff
@@ -114,5 +114,28 @@ Cloudant.prototype.getModelObjectSettings = function(mo) {
...
+      if (err) cb('ping failed');
+      self.getDriverInst().db.list(function(err, result) {
+        returnCB(err, result);
+      });
+    });
+  }
+
+  function returnCB(err, result) {
+    debug('Cloudant.prototype.ping results %j %j', err, result);
+    if (err) cb('ping failed');
```
b-admike: would `if (err) cb(new Error('ping failed'))` work better here?

jannyHou:
@kjdelisle we already have `ping` succeeds, for `ping` fails, other than an invalid `url`, I am not sure what I can add, any idea?

lib/cloudant.js
jannyHou: ah true, sorry I copied the old code

jannyHou:
@kjdelisle The test case is added in PR strongloop/loopback-connector-couchdb2#33, hope it passes CI and it should be checked in before this Cloudant PR, let's put the comment in that PR first :-p

lib/cloudant.js
```diff
@@ -114,5 +114,28 @@ Cloudant.prototype.getModelObjectSettings = function(mo) {
...
+  debug('Cloudant.prototype.ping');
+  var self = this;
+  var driverInst = self.getDriverInst();
+  if (driverInst) {
+    driverInst.db.list(function(err, result) {
+      returnCB(err, result);
+    });
+  } else {
+    self.dataSource.once('connected', function(err) {
+      if (err) cb('ping failed');
```
kjdelisle: `cb(new Error('ping failed'))`

kjdelisle: To match the same pattern below.

jannyHou: @kjdelisle Fixed :)

jannyHou:
@slnode test please

jannyHou:
@slnode test please

jannyHou:
@kjdelisle The ["fail" test case(checking `ping failed`)](https://github.com/strongloop/loopback-connector-couchdb2/blob/master/test/connection.test.js#L30-L31) won't work for cloudant, since event `connected` won't happen: https://github.com/strongloop/loopback-connector-cloudant/pull/177/files#diff-fa06ecf3249c63aabcbf7ff1ae8ec186R126

The error is 
> Error: connect ECONNREFUSED 127.0.0.1:4

instead of Error object `ping failed`

Are you ok if I disable that test case for cloudant? I don't think you can test it.

kjdelisle:
@jannyHou Are you saying it's a huge amount of effort to _override_ the test? If the expected response is not identical to the CouchDB response, it makes sense to reimplement the test to be Cloudant-specific, but I don't agree that we should just disable all forms of test coverage for the ping function.

jannyHou:
@kjdelisle I can write a test for it, but it's an uncaught error, not from cb

kjdelisle:
@jannyHou Is this a side-effect of the cloudant driver not correctly providing the error via callback/promise?

jannyHou:
@kjdelisle The uncaught is not even returned by `ping`, it gets generated when you do a `var db = getDataSource(fakeConfig)`, I don't think it's a driver's problem, the juggler is written in that way to return error, see test case in mongodb: https://github.com/strongloop/loopback-connector-mongodb/blob/master/test/mongodb.test.js#L35

The interesting part is I am not sure why the couchdb doesn't fail when calling `getDataSource` with a fake config.

kjdelisle:
@jannyHou Then that's the logic you should imitate for the cloudant version of the test.

kjdelisle:
@jannyHou As long as the tests pass (and are there!) I'm happy.

test/cloudant.connection.test.js
```diff
@@ -0,0 +1,24 @@
+// Copyright IBM Corp. 2017. All Rights Reserved.
+// Node module: loopback-connector-couchdb2
```
Amir-61: `loopback-connector-couchdb2` -> `loopback-connector-cloudant`

lib/cloudant.js
```diff
@@ -114,5 +114,28 @@ Cloudant.prototype.getModelObjectSettings = function(mo) {
...
+  if (driverInst) {
+    driverInst.db.list(function(err, result) {
+      returnCB(err, result);
+    });
+  } else {
+    self.dataSource.once('connected', function(err) {
+      if (err) cb(new Error('ping failed'));
+      self.getDriverInst().db.list(function(err, result) {
+        returnCB(err, result);
+      });
```
Amir-61:
```
      self.getDriverInst().db.list(function(err, result) {
         returnCB(err, result);
       });
```
can be changed to:
```
     self.getDriverInst().db.list(returnCB(err, result));
```
to me It is much cleaner and neater...

lib/cloudant.js
```diff
@@ -114,5 +114,28 @@ Cloudant.prototype.getModelObjectSettings = function(mo) {
...
 };
 
+Cloudant.prototype.ping = function(cb) {
+  debug('Cloudant.prototype.ping');
+  var self = this;
+  var driverInst = self.getDriverInst();
+  if (driverInst) {
+    driverInst.db.list(function(err, result) {
+      returnCB(err, result);
+    });
```
Amir-61:
```
    driverInst.db.list(function(err, result) {
       returnCB(err, result);
     });
```
can be changed to:
```
     driverInst.db.list(returnCB(err, result));
```
@kjdelisle @jannyHou to me It is much cleaner and neater...

test/cloudant.connection.test.js
kjdelisle: Missed that one, good catch

PR #184, opened 2018-01-31
Add global limit

jannyHou:
#### Related issues

https://github.com/strongloop/loopback-connector-couchdb2/issues/41

### Checklist


package.json
```diff
@@ -48,4 +51,4 @@
     "url": "https://github.com/strongloop/loopback-connector-cloudant/issues"
   },
   "license": "Artistic-2.0"
-}
+}
```
virkt25: Need newline

package.json
```diff
@@ -5,6 +5,9 @@
     "tag": "lts"
   },
   "description": "LoopBack Connector for IBM Cloundant DB",
+  "engines": {
+    "node": ">= 4 <= 6"
```
virkt25: Any reason this doesn't work with 8?

test/count.test.js
```diff
@@ -0,0 +1,57 @@
...
+// License text available at https://opensource.org/licenses/Apache-2.0
+
+'use strict';
+
+var _ = require('lodash');
+var should = require('should');
+var COUNT_OF_SAMPLES = 70;
+var db, TestCountUser;
+
+function create50Samples() {
```
virkt25: Please rename the function to just `createSamples`. It's misleading right now as it creates 70 samples but is called `create50Samples`

lib/cloudant.js
```diff
@@ -784,6 +786,15 @@ Cloudant.prototype.automigrate = function(models, cb) {
  *        "<key2>": -1
  *
 */
+
+Cloudant.prototype.getLimit = function(limit) {
+  return limit || this.getGlobalLimit();
+};
+
+Cloudant.prototype.getGlobalLimit = function() {
```
virkt25: Any reason for this to be a function and not just be a part of `getLimit`? 

```
Cloudant.prototype.getLimit = function(limit) {
  return limit || this.settings.globalLimit;
}
```

test/id.test.js
```diff
@@ -18,8 +18,7 @@ describe('using _id as id', function() {
       owner: {type: String},
       _id: {type: String, id: true},
     }, {forceId: true});
-
-    db.automigrate('Task', done);
```
virkt25: Is removing the `db.automigrate` intentional? (Seems unrelated to the issue -- just want to confirm it wasn't accidental).

package.json
jannyHou: @virkt25 This is branch 1.x and I copied the engines from `loopback-datasource-juggler@2.x`, I think they have the same engines support

test/count.test.js
jannyHou: good catch

test/id.test.js
jannyHou: @virkt25 1.x has a bug with `automigrate` function, which is fixed in 2.x, if you look at all the 1.x tests, they don't call `automigrate`, calling it results in a document conflict error.

lib/cloudant.js
jannyHou: It's more like a personal coding habit, I treat `getGlobalLimit` and `getLimit` as separate functionalities, e.g. you can add more approaches set the global limit value, currently it's from datasource settings only, but it may come from model or Env in the future.
And for `getLimit`, we may add options and corresponding logic.

test/id.test.js
jannyHou: find the related PR:
https://github.com/strongloop/loopback-connector-cloudant/commit/e348c2576d7fead66ce7370871fe5cfecef56539
and 
https://github.com/strongloop/loopback-connector-cloudant/pull/109
if you are interested.

package.json
jannyHou: turned off `autosave` :(

test/id.test.js
virkt25: Ooh ok. I just wanted to make sure it wasn't accidental and I didn't realize this was for 1.x branch

package.json
virkt25: `1.x` can still be used with `3.x` version of Juggler. I disagree with it only supporting what 2.x supports because 3.x does support Node 8. Unless this connector version of 1.x shouldn't be used with Juggler 3.x at all.

lib/cloudant.js
virkt25: Fair enough. My personal preference is just having one function :P 

This isn't a blocker.

package.json
jannyHou: The code of `1.x` is written when node@8.x was not released, and due to the annoying bug of `automigrate`, I would not recommend new users to use it, `1.x` doesn't have the index and view feature either. That's why I don't want to spend effort on its compatibility of node 8, even it passes all test on node 8, I still don't want to give loopback3 users an impression that this old version is an option for them :(

jannyHou:
@virkt25 thanks for the review, good talk. green now :)

PR #192, opened 2018-06-19
chore: update node versions in travis

dhmlau:
### Description
Drop Node 4, and add Node 8 and 10 in Travis.


dhmlau:
@b-admike , thanks for the reminder!  commit message is updated.

PR #186, opened 2018-02-15
Updated cloudant dependency

Danwakeem:
### Description
I simply updated the cloudant dependency because there is something wrong with this version of cloudant in the npm registry. 

I went ahead and added a PR for this because we are using this package in some of our apps that run on bluemix and our deployment pipelines are failing because this package will not build and I figured there may be other people having the same issue. :) 

Just FYI for any one stumbling across this, we have temporarily fixed this deployment issue by pointing our npm packages to this fork. But we are hoping to point it back to the main repo as soon as we can. 

#### Related issues
- connect to #185 
### Checklist

- [ ] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style
  guide](http://loopback.io/doc/en/contrib/style-guide.html)

PR #195, opened 2018-09-19
update cloudant

jannyHou:
### Description

Same as https://github.com/strongloop/loopback-connector-cloudant/pull/186
#### Related issues

- connect to <link_to_referenced_issue>

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
   "dependencies": {
     "async": "^1.5.0",
     "bluebird": "^3.1.1",
-    "cloudant": "^1.4.0",
+    "cloudant": "^1.10.0-NOTICE",
     "debug": "^2.2.0",
     "lodash": "^3.10.1",
     "loopback-connector": "^2.3.0",


package.json
```diff
@@ -25,7 +25,7 @@
   "dependencies": {
     "async": "^1.5.0",
     "bluebird": "^3.1.1",
-    "cloudant": "^1.4.0",
+    "cloudant": "^1.10.0-NOTICE",
```
virkt25: Should we use `@cloudant/cloudant`? And why are we using a version tagged with `-NOTICE`?

jannyHou: @virkt25 I copied from the PR that fixed the installation problem and also tried on local "^1.10.0-NOTICE" works.
`@cloudant/cloudant` may contain breaking change that we are not compatible with. If time allows we can try upgrading it in `loopback-connector-cloudant@2.x`, but for `loopback-connector-cloudant@1.x` I would rather stay with 1.x...

jannyHou: @virkt25 see this PR https://github.com/strongloop/loopback-connector-cloudant/pull/186

PR #199, opened 2019-02-12
chore: update strong-globalize

dhmlau:
### Description
Update `strong-globalize` version to use the latest 4.x version, and other dependencies.

Related to https://github.com/strongloop-internal/scrum-apex/issues/372


dhmlau:
@b-admike , commit message updated per ur request.

PR #206, opened 2019-06-17
juggler-v3-v4 shared test

agnes512:
### Description
At the moment, our connectors are installing juggler 3.x to run the test suite. When we make a change to juggler@3, cis-jenkins detects downstream dependency and triggers CI build of connectors with the modified juggler version. When we make a change to juggler's master, no downstream builds are triggered.

In the past, we were maintaining multiple branches to test connectors against different juggler versions. E.g. loopback-2.x containing code from "master" but installing juggler 2.x for testing. This is rather impractical, because we don't have bandwidth to update those other branches with changes made to master.

In this pull request, I am reworking connector test setup to execute tests from both juggler versions 3.x and 4.x, which are triggered in `./test.js`.

Drop `Node.js 6` CI tests as Juggler v4 does not support it. Also update `eslint` to the latest version (see those 2 PRs below).

#### Related issues
- Spike story: strongloop-internal/scrum-apex#416
- See also: strongloop/loopback-connector-postgresql#376

#### References issues
- https://github.com/strongloop/loopback-connector-mongodb/pull/519/commits/830f83a9eb59190975436b9d7bf353fe743b219b
- https://github.com/strongloop/loopback-connector-kv-redis/pull/43
- blocked by https://github.com/strongloop/loopback-datasource-juggler/pull/1750. Once the latest version gets released, the CI should be fine.

- resolves to https://github.com/strongloop/loopback-connector-cloudant/issues/205

### Checklist

- [ ] New tests added or existing tests modified to cover all changes
- [ ] Code conforms with the [style
  guide](http://loopback.io/doc/en/contrib/style-guide.html)

PR #207, opened 2019-06-20
update node version to skip imported test for older version

agnes512:
### Description
As we are changing to execute tests from both juggler versions 3.x and 4.x, and juggler v4 does not support Node.js 6, update node configs to skip tests for Node.js 6.

#### Related issues

- connect to https://github.com/strongloop/loopback-connector-cloudant/pull/206

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "2.3.0",
   "description": "LoopBack Connector for IBM Cloudant DB",
   "engines": {
-    "node": ">=6"
+    "node": ">=8"
   },
   "author": "IBM Corp.",
   "keywords": [
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,5 @@ before_install:
   - npm config set registry http://ci.strongloop.com:4873/
 
 node_js:
-  - 6
   - 8
   - 10


package.json
```diff
@@ -3,7 +3,7 @@
   "version": "2.3.0",
   "description": "LoopBack Connector for IBM Cloudant DB",
   "engines": {
-    "node": ">=6"
+    "node": ">=8.9"
```
jannyHou: "node": ">=8"

nabdelgadir: another connector does https://github.com/strongloop/loopback-connector-mongodb/blob/master/package.json#L6 but loopback-next has `8.9` but do you think we should follow other connectors?

agnes512: @jannyHou

jannyHou: @nabdelgadir @agnes512 Usually "8" is good, loopback-next restricts it to be above "8.9" because lower 8 versions don't work....But for connectors we don't have that problem.

package.json
```diff
@@ -3,7 +3,7 @@
   "version": "2.3.0",
   "description": "LoopBack Connector for IBM Cloudant DB",
   "engines": {
-    "node": ">=6"
+    "node": ">=8"
```
bajtos: Let's use `8.9` please. Older 8.x versions of Node are missing few useful features we are depending on in LB4.

bajtos: Never mind, I see you have discussed this aspect, I am fine with your decision.

PR #208, opened 2019-06-21
Update eslint version

agnes512:
### Description
As we are changing to execute tests from both juggler versions 3.x and 4.x, update `eslint` to newer version to pass CI tests.
The version of `eslint` matches with ~[mongoDB connector](https://github.com/strongloop/loopback-connector-mongodb/commit/830f83a9eb59190975436b9d7bf353fe743b219b#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R40)~ latest version

#### Related issues

- connect to https://github.com/strongloop/loopback-connector-cloudant/pull/206

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
   "main": "index.js",
   "scripts": {
     "lint": "eslint .",
+    "lint:fix": "npm run lint -- --fix",
     "test": "node test.js",
     "posttest": "npm run lint",
     "mocha": "./node_modules/.bin/_mocha --timeout 40000 --require strong-mocha-interfaces --require test/init.js --ui strong-bdd"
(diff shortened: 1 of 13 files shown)


package.json
```diff
@@ -33,8 +34,8 @@
   },
   "devDependencies": {
     "dockerode": "^2.4.3",
-    "eslint": "^2.13.1",
-    "eslint-config-loopback": "^4.0.0",
+    "eslint": "^5.1.0",
+    "eslint-config-loopback": "^10.0.0",
```
nabdelgadir: is there a reason you're using these versions instead of the latest, just curious?

agnes512: I wanted to match it with [another connector]( https://github.com/strongloop/loopback-connector-mongodb/commit/830f83a9eb59190975436b9d7bf353fe743b219b#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R40)

bajtos: It's best to upgrade to the latest version. Other connectors should be eventually upgraded too.

PR #223, opened 2019-10-18
update docker setup script

agnes512:
### Description

connect to https://github.com/strongloop/loopback-next/pull/3968

The current script failed to create new database:
```
curl --request PUT --url http://$USER:$PASSWORD@$HOST:$PORT/$DATABASE > /dev/null 2>&1
```

Update the script so that we can run acceptance tests in `repository-tests` on cloudant.

### Checklist

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,18 +1,17 @@
-version: '2'
+version: "3"
 services:
   cloudant:
     image: ibmcom/cloudant-developer:latest
     container_name: cloudant-testdb
     restart: always
     privileged: true
-    hostname: cloudant
+    hostname: testdb.local
     ports:
       - "8080:80"
     expose:
       - "80"
     volumes:
       - cloudant-testdb:/srv
-    mem_limit: 256M
     network_mode: bridge
 volumes:
-    cloudant-testdb:
+  cloudant-testdb:
--- a/setup.sh
+++ b/setup.sh
@@ -105,7 +105,7 @@ printf "\n${CYAN}Cloudant started.${PLAIN}\n"
 
 ## create database		
 printf "\n${RED}>> Creating database in Cloudant${PLAIN}"		
-curl --request PUT --url http://$USER:$PASSWORD@$HOST:$PORT/$DATABASE > /dev/null 2>&1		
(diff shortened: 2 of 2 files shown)


setup.sh
```diff
@@ -105,7 +105,7 @@ printf "\n${CYAN}Cloudant started.${PLAIN}\n"
 
 ## create database		
 printf "\n${RED}>> Creating database in Cloudant${PLAIN}"		
-curl --request PUT --url http://$USER:$PASSWORD@$HOST:$PORT/$DATABASE > /dev/null 2>&1		
+curl -i -X PUT http://$USER:$PASSWORD@$HOST:$PORT/$DATABASE > /dev/null 2>&1
```
bajtos: Can we remove `2>&1`, so that when the request fails, the errors are printed by the `setup.sh` script?

I also believe that `-i` is not necessary, since you are discarding the output to `/dev/null` anyways.

Quoting from `man curl`:

> -i, --include
> Include the HTTP-header in the output. The HTTP-header includes things like server-name, date of the document, HTTP-version and more...
>
> See also -v, --verbose.

Proposed line:


```suggestion
curl -X PUT http://$USER:$PASSWORD@$HOST:$PORT/$DATABASE > /dev/null	
```

emonddr: npm run mocha
npm run test 

fail with some errors on my laptop, but works in CI build?

emonddr: hmm some tests fail on laptop with master branch too.

Why docker-compose use the `:latest` tag, but `test.js` use `:2.0.1` ?

emonddr: seem to be the same

image

docker-compose.yml
```diff
@@ -1,18 +1,17 @@
-version: '2'
+version: "3"
 services:
   cloudant:
     image: ibmcom/cloudant-developer:latest
```
jannyHou: This image is retired, I would suggest we use `ibmcom/couchdb3` instead, see https://hub.docker.com/r/ibmcom/couchdb3.

agnes512:
> We need to find out why

It failed before the changes.

agnes512:
closing as we've decided to use couchdb to run the test.

emonddr:
@agnes512 , can you delete this branch now? Since this PR is not going to merge? thx.

Issue #215, opened 2019-10-07
Update driver to nodejs/cloudant


jannyHou:
See parent spike story https://github.com/strongloop/loopback-connector-cloudant/issues/214

### Acceptance Criteria 

- [ ] Update driver to https://github.com/cloudant/nodejs-cloudant
- [ ] Find a new db (support partition)for test
    - The docker one is retired and don't support partition
    - Check the couchdb2 image as well, and replace it with a new one if necessary. (since 95% of the code are written in loopback-connector-couchdb2)

jannyHou:
The new docker image: https://hub.docker.com/r/ibmcom/couchdb3

emonddr:
@jannyHou have you submitted a PR for this yet? Is this possible for October?

jannyHou:
@emonddr I have code change on local, haven't created the PR. I see a PR updating the docker image https://github.com/strongloop/loopback-connector-cloudant/pull/223#pullrequestreview-303991999, it would be better if we can upgrade to use https://hub.docker.com/r/ibmcom/couchdb3.

emonddr:
> #223 (review)

@agnes512  closed this PR after we had a discussion, and she realized it wasn't needed.

jannyHou:
closed as https://github.com/strongloop/loopback-connector-cloudant/pull/224 landed

PR #224, opened 2019-10-21
update docker image

jannyHou:
### Description

- Replace the docker image with new one 
- Update the driver to @cloudant/cloudant

For the skipped geo query test, see my comment in https://github.com/strongloop/loopback-connector-cloudant/pull/224/files#r337751334

#### Related issues

https://github.com/strongloop/loopback-connector-cloudant/issues/215

### Checklist


docker-compose.yml
```diff
@@ -1,15 +1,15 @@
 version: '2'
 services:
   cloudant:
-    image: ibmcom/cloudant-developer:latest
+    image: ibmcom/couchdb3:latest
     container_name: cloudant-testdb
```
agnes512: so even we use couchdb's image, we still name the container `cloudant-testdb`?

jannyHou: @agnes512 yeah because we have another connector in https://github.com/strongloop/loopback-connector-couchdb2, we need to distinguish them.

test/init.js
```diff
@@ -32,6 +32,7 @@ const skips = [
   'find all skip two hundred',
   'isActual',
   'with an invalid connection',
+  'cloudant geo',
```
jannyHou: The test is skipped because geo query not working as before.
I updated the test case according to https://github.com/cloudant/nodejs-cloudant#cloudant-geospatial, and realized the example doesn't return the expected result. The query is rejected with "Document is missing" error.
Gist https://gist.github.com/jannyHou/f865a1cb9108c3848e63f97177f58030#file-gistfile1-js-L61 reproduces it, confirming with cloudant team on slack, will create an issue in https://github.com/cloudant/nodejs-cloudant if it's a bug.

jannyHou:
@slnode test please

jannyHou:
@slnode test please

jannyHou:
@raymondfeng The docker compose file is used for local setup. CI doesn't use it. `setup.sh` and `docker-compose.yaml` are both for running tests on local.

test/cloudant.test.js
```diff
@@ -498,20 +505,6 @@ describe('cloudant constructor', function() {
     done();
   });
 
-  it('should give 401 error for wrong creds', function(done) {
```
jannyHou: a dup test case of https://github.com/strongloop/loopback-connector-cloudant/blob/master/test/cloudant.connection.test.js

Issue #216, opened 2019-10-07
Support creating partitioned index


jannyHou:
See parent story https://github.com/strongloop/loopback-connector-cloudant/issues/214

### Acceptance Criteria

- [x] Add configurable property `partitioned` for index support
  - By default an index is created with `{partitioned: false}` as global.
  - When people define the index in the model definition, they can configure the partitioned as `true/false`.

jannyHou:
Closing after https://github.com/strongloop/loopback-connector-cloudant/pull/225 merged.

PR #225, opened 2019-11-09
feat: partitioned index

jannyHou:
### Description

connect to https://github.com/strongloop/loopback-connector-cloudant/issues/216
Support create partitioned index in partitioned database.

User can define the index in model config as 
```js
Product = db.define('Product', {
      prodName: {type: String},
      desc: {type: String},
    }, {
      indexes: {
        'prodName_index': {
         // Partitioned index
          partitioned: true,
          keys: {
            prodName: 1,
          },
        },
      },
    });
```
to create partitioned index.
If no `partitioned` property defined, the default index will be global. 

~TODO:~

- [x] fix lint
- [x] add jsdoc
- [x] extract util method as prototype private method
- [x] refactor couchdb2, PR https://github.com/strongloop/loopback-connector-couchdb2/pull/69 to be merged
- [x] delete dup function in cloudant

#### Related issues

- connect to <link_to_referenced_issue>

### Checklist

### DIFF
--- a/lib/migrate.js
+++ b/lib/migrate.js
@@ -1,0 +1,207 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: loopback-connector-cloudant
+// This file is licensed under the Apache License 2.0.
+// License text available at https://opensource.org/licenses/Apache-2.0
+
+'use strict';
+
+const _ = require('lodash');
+const inspect = require('util').inspect;
+const debug = require('debug')('loopback:connector:cloudant');
+
+module.exports = mixinMigrate;
+
+function mixinMigrate(Cloudant) {
+  /**
+    * Used in function `migrateOrUpdateIndex`.
+    * Create index with index object.
+    * @param {Object} mo  The model configuration
+    * @param {String} model  The model name.
+    * @param {String} name  The index name.
+    * @param {Object} indexObj The index object
+    * e.g. {fields: [{"foo": "asc"}, {"bar": "asc"}], partitioned: true}
+    * @param {Function} cb
+  */
+  Cloudant.prototype._createIndex = function(mo, model, name, indexObj, cb) {
+    const self = this;
+    debug('createIndex fields before coercion: %s', inspect(indexObj.fields));
+    let fields = self.coerceIndexFields(indexObj.fields);
+    debug('createIndex fields after coercion: %s', inspect(fields));
+    fields = self.addModelViewToIndex(mo.modelView, fields);
+    debug('createIndex fields after add model index: %s', inspect(fields));
+    // naming convertion: '_design/LBModel__Foo__LBIndex__foo_index',
+    // here the driver api takes in the name without prefix '_design/'
+    const config = {
+      ddocName: self.getIndexModelPrefix(mo) + '__' + model + '__' +
+      self.getIndexPropertyPrefix(mo) + '__' + name,
+      indexName: name,
+      fields: fields,
+      partitioned: indexObj.partitioned,
+    };
+    self.createIndexWithConfig(config, cb);
+  };
+
+  /**
+   * Used in function `getModifyIndexes()`.
+   * Generate indexes for model properties that are configured as
+   * `{index: true}`, return model property indexes as objects,
+   * e.g. {fields: [{name: 'asc'}], partitioned: true}.
+   *
+   * @param {Object} indexes indexes from model config, retrieved in
+   * `getModifyIndexes()`
+   */
+  Cloudant.prototype._generatePropertyLevelIndexes = function(indexes) {
+    const results = {};
+    for (const key in indexes) {
+      const field = {};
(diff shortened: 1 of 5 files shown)


lib/migrate.js
```diff
@@ -0,0 +1,207 @@
...
+    * Used in function `migrateOrUpdateIndex`.
+    * Create index with index object.
+    * @param {Object} mo  The model configuration
+    * @param {String} model  The model name.
+    * @param {String} name  The index name.
+    * @param {Object} indexObj The index object
+    * e.g. {fields: [{"foo": "asc"}, {"bar": "asc"}], partitioned: true}
+    * @param {Function} cb
+  */
+  Cloudant.prototype.__createIndex = function(mo, model, name, indexObj, cb) {
```
dhmlau: similar to [comment](https://github.com/strongloop/loopback-connector-couchdb2/pull/69#discussion_r345284015) from @bajtos, it should probably be single `_` in the function name.

jannyHou: 👍 renamed them to single `_`

lib/migrate.js
```diff
@@ -0,0 +1,207 @@
...
+        obj[key] = order;
+        fields.push(obj);
+      });
+      results[key] = {fields, partitioned};
+    }
+    return results;
+  };
+
+  /**
+ * Perform the indexes comparison for `autoupdate`.
```
agnes512: Is this function for comparing if indexes exist?

test/partition.test.js
```diff
@@ -0,0 +1,164 @@
...
+        index.def.fields[0]['prodName'].should.equal('desc');
+        index.def.fields[1][DEFAULT_MODEL_VIEW].should.equal('desc');
+        // should be a global index
+        index.partitioned.should.equal(false);
+        done();
+      });
+    });
+  });
+
+  it('configured true - create partitioned index for index entry', (done) => {
```
agnes512: I feel like a test title like `create partitioned index for index entry with partition enabled` would make the intention more clear. Same for `configured false` test. ( Unless the option we enable/disable `partitioned` called `configured`)

test/partition.test.js
```diff
@@ -0,0 +1,164 @@
...
+        const indexName = 'prodName1_index';
+        should.exist(indexes);
+
+        const index = _.find(indexes, function(index) {
+          return index.name === indexName;
+        });
+        should.exist(index);
+        index.name.should.equal(indexName);
+        index.def.fields[0]['prodName'].should.equal('desc');
+        index.def.fields[1][DEFAULT_MODEL_VIEW].should.equal('desc');
```
agnes512: I am not familiar with Cloudant index field. Why is this `desc` and others are `asc`?

lib/migrate.js
jannyHou: @agnes512 yep, it returns the new indexes to add and old indexes to drop

test/partition.test.js
jannyHou: @agnes512 Good question, see https://github.com/strongloop/loopback-connector-cloudant/pull/225/files#diff-121c1d00e6380aa6109385e1df1ae803R66

The order code specified for `prodName` is -1, which stands for `desc`

test/partition.test.js
jannyHou: renamed all the tests title :)

test/partition.test.js
agnes512: looks like they need a better name :laughing:

test.js
```diff
@@ -37,7 +38,8 @@ async.waterfall([
   setCloudantEnv,
   waitFor('/_all_dbs'),
   createAdmin(),
-  createDB('test-db'),
+  createDB(process.env.CLOUDANT_DATABASE),
+  createPartitionedDB(process.env.CLOUDANT_PARTITIONED_DATABASE),
```
dhmlau: this doesn't seem to be called in other places?

test/partition.test.js
```diff
@@ -0,0 +1,168 @@
...
+      });
+    });
+  });
+
+  it('index entry - create global index by default', (done) => {
+    Product = db.define('Product', {
+      prodName: {type: String},
+      desc: {type: String},
+    }, {
+      indexes: {
```
dhmlau: not really about this test case.. Do we have any tests to show after we've created a database using the `partitioned:true` option, the `indexes` look like this?

test.js
dhmlau: @jannyHou showed me where it's being used.

test/partition.test.js
dhmlau: Discussed it with @jannyHou. This test case is about if someone specify an `indexes` for the product, it should be preserved after putting it into the database. 
For the `partitioned:true` "property", it is at the global level, so if we don't specify any `indexes`, there will be no difference for the indexes between a partitioned database and a non-partitioned database.

Issue #220, opened 2019-10-08
findAll: Read partitionKey from `options`


jannyHou:
See parent story: https://github.com/strongloop/loopback-connector-cloudant/issues/214

## Acceptance Criteria

- [x] For method `Couchdb2.prototype.all()`(which maps to model method `find`), get the partition key from `options` like: 
  ```js
   Customer.find({
      where: { name: 'somename' }
   },
   {
      // options
      partitonKey: 'US'
   });
   ```
   If `partitionKey` is found, then invoke `driverInst.partitionedFind(partitionKey, query)` to find documents.

raymondfeng:
It should be `{cloudant: {partitionKey: 'US'}}` or `{partitionKey: 'US'}`.

jannyHou:
Closing since PR merged

PR #229, opened 2019-11-18
Partition key from option

jannyHou:
### Description

Prerequisite PR: https://github.com/strongloop/loopback-connector-couchdb2/pull/71

connect to https://github.com/strongloop/loopback-connector-cloudant/issues/220

To trigger the `partitonedFind`, you can include the `partitionKey` field in `options` like:

```
Product.find(
  {where: {tag: 'food'}},
  // specify the partition key in the options
  {partitionKey: 'toronto'},
  cb
)
```

#### Related issues

- connect to https://github.com/strongloop/loopback-connector-cloudant/issues/220

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,9 +29,10 @@
     "fs-extra": "^8.1.0",
     "lodash": "^4.17.11",
     "loopback-connector": "^4.0.0",
-    "loopback-connector-couchdb2": "^1.5.1",
+    "loopback-connector-couchdb2": "^1.5.2",
     "request": "^2.81.0",
-    "strong-globalize": "^5.0.0"
+    "strong-globalize": "^5.0.0",
+    "uuid": "^3.3.3"
   },
   "devDependencies": {
     "dockerode": "^2.4.3",
--- a/test/partition.test.js
+++ b/test/partition.test.js
@@ -8,6 +8,7 @@
 require('./init.js');
 const _ = require('lodash');
 const should = require('should');
+const uuid = require('uuid/v4');
 const DEFAULT_MODEL_VIEW = 'loopback__model__name';
 let Product, db, connector;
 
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -29,9 +29,10 @@
     "fs-extra": "^8.1.0",
     "lodash": "^4.17.11",
     "loopback-connector": "^4.0.0",
-    "loopback-connector-couchdb2": "^1.5.1",
+    "loopback-connector-couchdb2": "github:strongloop/loopback-connector-couchdb2#add-option-for-search",
```
raymondfeng: Don't forget to switch back to npm released version.

jannyHou: 👍 sure

test/partition.test.js
```diff
@@ -28,141 +29,188 @@ describe('cloudant - partitioned db', () => {
...
+            keys: {
+              city: 1,
+            },
+          },
+        },
+      });
+      db.automigrate('Product', () => {
+        Product.create(SEED_DATA, function(err) {
+          if (err) return done(err);
+          Product.find({where: {tag: 'food'}}, {partitionKey: 'toronto'},
```
agnes512: Just out of curious, I guess the performance of `partition` + `where`(#L196) is better than doing `where` on `find()` normally?

agnes512: Just out of curious, I guess the performance of partition + where(#L196) is better than doing where on find() normally?

agnes512: Just out of curious, I guess the performance of partition + where(#L196) is better than doing where on find() normally?

test/partition.test.js
```diff
@@ -28,141 +29,188 @@ describe('cloudant - partitioned db', () => {
...
 });
+
+const SEED_DATA = [
+  {id: `toronto: ${uuid()}`, name: 'beer', tag: 'drink'},
+  {id: `toronto: ${uuid()}`, name: 'salad', tag: 'food'},
+  {id: `toronto: ${uuid()}`, name: 'soup', tag: 'food'},
+  {id: `london: ${uuid()}`, name: 'beer', tag: 'drink'},
+  {id: `london: ${uuid()}`, name: 'salad', tag: 'food'},
+  {id: `london: ${uuid()}`, name: 'salad', tag: 'food'},
+];
```
agnes512: What if I have a 7th record `{id: toronto: ${uuid()}, name: 'dessert', tag: 'food'},`, would it be included to the result of the above test case?

test/partition.test.js
jannyHou: @agnes512 Exactly, if you are interested to learn more about this feature, it's explained in https://docs.couchdb.org/en/master/partitioned-dbs/index.html.

An example: you have 200 documents in partition `US` and 300 documents in partition `CA`, when you do a global search, you find among 500 documents, but if you specify `partitionKey` as `US` and perform a partitioned search, you find among those 200 documents.

test/partition.test.js
jannyHou: yep, because its partitionKey is `toronto`(matches your partition key) and is tagged as `food`(match your query).

agnes512: I see, the name `partitionKey` always makes me think of operations based on index.

jannyHou: Yep they are related :) 
> The benefit of partitioned databases is that secondary indices can be significantly more efficient when locating matching documents since their entries are contained within their partition. This means a given secondary index read will only scan a single partition range instead of having to read from a copy of every shard.

My understanding is index is a secondary optimization for search.

Issue #217, opened 2019-10-07
Optimize query by reading partition key from payload


jannyHou:
See parent story https://github.com/strongloop/loopback-connector-cloudant/issues/214

### Acceptance Criteria 

Model `Order` has `userId` as its partition key.

- [x] Support providing the `partitionKey` in query, like `Order.find({where: {somefield: 'somevalue', userId: 'auser',}})`

    - In `Cloudant.prototype.find()`, if `partitionKey` is detected in the where clause, then invoke `db.partitionedFind('auser', queryWithPartitionKeyExcluded)`
    - Please note the `partitionKey` from `options` should override the one from where clause.

jannyHou:
closing as PR merged

PR #230, opened 2019-11-20
Partition field

jannyHou:
### Description

connect to #217 

- You can specify a partition key field in model definition like
```js
customers = db.define('customers', {
  userId: {type: Number, id: true},
  // partition key field
  countryCode: {type: String, isPartitionKey: true},
  name: String,
  zipCode: Number,
});
```

PR https://github.com/strongloop/loopback-connector-couchdb2/pull/72 adds the name of partition key field to couchdb/cloudant's model object. 

- When the partition key field is included in query, `Cloudant.prototype.all` will trigger parititonFind from driver instead of a global search.

- The partition key provided in `options` will override the one in query.

#### Related issues

- connect to #217

### Checklist


test/partition.test.js
```diff
@@ -185,32 +186,60 @@ describe('cloudant - partitioned db', () => {
...
-                });
-              resultTaggedFood.length.should.equal(2);
-              done();
+          should.exist(results);
+          results.length.should.equal(2);
+          const resultTaggedFood = _.filter(results,
+            function(r) {
+              return r.tag === 'food' && r.city === 'toronto';
             });
+          resultTaggedFood.length.should.equal(2);
```
agnes512: Why doesn't it equal to 3?

test/partition.test.js
```diff
@@ -185,32 +186,60 @@ describe('cloudant - partitioned db', () => {
...
+          done();
+        });
+    });
+
+    it('find all records by partition key from property', (done) => {
+      Product.find({where: {tag: 'food', city: 'toronto'}},
+        (err, results) => {
+          if (err) return done(err);
+          should.exist(results);
+          results.length.should.equal(2);
```
agnes512: Why doesn't it equal to 3? and what's the difference between `find all records by partition key from option` and `find all records by partition key from property`? Does it mean we can invoke the partition key from property or option, and they they give the same result?

test/partition.test.js
```diff
@@ -185,32 +186,60 @@ describe('cloudant - partitioned db', () => {
...
+          const resultTaggedFood = _.filter(results,
+            function(r) {
+              return r.tag === 'food' && r.city === 'toronto';
+            });
+          resultTaggedFood.length.should.equal(2);
+          done();
+        });
+    });
+
+    it('partition key in options overrides property for find all', (done) => {
```
agnes512: Do you mean the `partitionKey` in option changes `city: {type: String, isPartitionKey: true}` to `city: {type: String, isPartitionKey: false}`?

test/partition.test.js
```diff
@@ -185,32 +186,60 @@ describe('cloudant - partitioned db', () => {
...
+      Product.find({where: {tag: 'food', city: 'toronto'}},
+        {partitionKey: 'london'}, (err, results) => {
+          if (err) return done(err);
+          should.exist(results);
+          results.length.should.equal(0);
```
agnes512: Do you mean the options overrides the query clause? 

And if it does, why isn't it result to 2? ( `food` and `london`)

jannyHou: @agnes512 
> Do you mean the options overrides the query clause?

Yes, it's 0 because when search within partition `london`, there is no record with `city` as `toronto`. See the sample records created in https://github.com/strongloop/loopback-connector-cloudant/pull/230/files#diff-121c1d00e6380aa6109385e1df1ae803R237

test/partition.test.js
jannyHou: Seems the tests are confusing, let me add more docs for it.

lib/cloudant.js
```diff
@@ -134,6 +134,29 @@ Cloudant.prototype.ping = function(cb) {
...
+ *
+ * @param {Object} mo The model object with properties consumed
+ * by cloudant connector, generated by `CouchDB.prototype.selectModel`.
+ * @param {Object} query The cloudant syntax query built from request query.
+ * @param {Object} options The options from caller.
+ */
+Cloudant.prototype._getPartitionKey = function(mo, query, options) {
+  const pkFromOption = options && options.partitionKey;
+  debug('_getPartitionKey pkFromOption %s', pkFromOption);
+  if (pkFromOption) return pkFromOption;
```
jannyHou: if option has partitionKey, return immediately.

test/partition.test.js
```diff
@@ -223,6 +229,10 @@ describe('cloudant - partitioned db', () => {
     });
 
     it('partition key in options overrides property for find all', (done) => {
+      // Find records with partition `toronto`, tag `food`, city `toronto`,
```
agnes512: with partition `toronto` -> with partition `london`?

test/partition.test.js
```diff
@@ -235,11 +245,18 @@ describe('cloudant - partitioned db', () => {
 });
 
 const SEED_DATA = [
+  // partition `toronto`, tag `drink`
```
agnes512: 👍 the comments are helpful

test/partition.test.js
jannyHou: good catch !

PR #240, opened 2020-03-12
Exclude 'deps' and '.github' from npm publish

emonddr:
Related to: https://github.com/strongloop/loopback-connector-mssql/issues/222 

Using similar fix applied to loopback-connector-mssql : https://github.com/strongloop/loopback-connector-mssql/pull/226 

## Checklist

👉 [Read and sign the CLA (Contributor License Agreement)](https://cla.strongloop.com/agreements/strongloop/loopback-connector-cloudant) 👈

- [x] `npm test` passes on your machine
- [ ] Code conforms with the [style guide](https://loopback.io/doc/en/contrib/style-guide-es6.html)
- [ ] Commit messages are following our [guidelines](https://loopback.io/doc/en/contrib/git-commit-messages.html)

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,6 +22,13 @@
     "posttest": "npm run lint",
     "mocha": "./node_modules/.bin/_mocha --timeout 40000 --require strong-mocha-interfaces --require test/init.js --ui strong-bdd"
   },
+  "files": [
+    "intl",
+    "lib",
+    "docs.json",
+    "index.js",
+    "setup.sh"
+  ],
   "dependencies": {
     "@cloudant/cloudant": "^4.2.1",
     "async": "^3.1.0",
--- a/.npmignore
+++ b/.npmignore
@@ -1,5 +1,0 @@
-node_modules
-.DS_Store
-npm-debug.log
-test
-example


package.json
```diff
@@ -22,6 +22,13 @@
     "posttest": "npm run lint",
     "mocha": "./node_modules/.bin/_mocha --timeout 40000 --require strong-mocha-interfaces --require test/init.js --ui strong-bdd"
   },
+  "files": [
+    "intl",
+    "lib",
+    "docs.json",
+    "index.js",
+    "setup.sh"
```
jannyHou: cloudant contains more setup files I think:
setup.sh
test.setup.sh
test.js
docker.setup.js
docker-compose.yaml

jannyHou: on a second thought, I am not sure do we need to include the setup files when install the connector as a dependency. usually the app should provide a valid db. So probably remove the `setup.sh` file?

jannyHou: back from discussion in 
I see why `setup.sh` is needed in other connectors.
Tip: lb-next/repository-tests has standalone setup script for cloudant, see https://github.com/strongloop/loopback-next/blob/master/acceptance/repository-cloudant/setup.sh, so including/excluding it in `files` doesn't affect the tests. I will honor @emonddr 's decision on it.

emonddr: > back from discussion in https://github.com/strongloop/loopback-connector-mysql/pull/415/files#r391873007
> I see why `setup.sh` is needed in other connectors.
> Tip: lb-next/repository-tests has standalone setup script for cloudant, see https://github.com/strongloop/loopback-next/blob/master/acceptance/repository-cloudant/setup.sh, so including/excluding it in `files` doesn't affect the tests. I will honor @emonddr 's decision on it.

I figure we should be consistent in what we include in "files" section.

Issue #213, opened 2019-08-05
FindById doesn't use Cloudant Read by Id method


mgzeitouni:
The FindById is not actually using Cloudant's Id Lookup method ('Read'), rather it uses a standard query (_find), which is 10x more expensive. Is there an option that uses a Read for the findById function?

dhmlau:
@jannyHou, I remember there was a discussion about the cost of calling lookup vs query for Cloudant, but I cannot recall you had any workaround or solution.  Could you please help? Thanks.

dhmlau:
@mgzeitouni, sorry about the late response. 
From my understanding, @jannyHou had made some enhancement for `findById` at the connector level.  

I found her response in another thread: 
> there is a way to do fetch id: call function `ds.connector.findById(modelName, id, options, cb)`. 
> See function https://github.com/strongloop/loopback-connector-couchdb2/blob/master/lib/couchdb.js#L482
> You need to call it as a datasource method instead of a model method, please check the doc for function viewDocs to see how to call a datasource method:
> https://github.com/strongloop/loopback-connector-couchdb2#view 

Ideally we can make this available to the juggler level, however, the effort is huge.  Please let us know if the above comment helps.

Making use of the partitioned database feature is another way to reduce the cost of using the cloudant service.  Please see the discussion in this epic: https://github.com/strongloop/loopback-connector-cloudant/issues/219.  We hope to finish the MVP (i.e. described in #219) in Q4 and make other enhancement in the post MVP epic https://github.com/strongloop/loopback-connector-cloudant/issues/222 as the next step.  I'd encourage you to review the 2 epics and continue the discussion or provide your inputs over there.  Also, contribution is always welcomed! :)

dhmlau:
@jannyHou, in case i miss anything, please chime in!

jannyHou:
@dhmlau Thank you for posting the reply here! That pretty much covers all the discussion 👍

bajtos:
I feel this issue is not fully resolved yet, I'd like to keep it open until we have a clear and well documented recommendation explaining LoopBack 4 users how to get an efficient `findById` implementation using Cloudant's Read operation.

PR #241, opened 2020-05-14
docs: findbyid

jannyHou:
Add doc for find by id

Connect to #213 

## Checklist

👉 [Read and sign the CLA (Contributor License Agreement)](https://cla.strongloop.com/agreements/strongloop/loopback-connector-cloudant) 👈

- [ ] `npm test` passes on your machine
- [ ] Code conforms with the [style guide](https://loopback.io/doc/en/contrib/style-guide-es6.html)
- [ ] Commit messages are following our [guidelines](https://loopback.io/doc/en/contrib/git-commit-messages.html)

### DIFF
--- a/README.md
+++ b/README.md
@@ -633,6 +633,10 @@ We are still in progress of refactoring some methods, more details to be updated
 
 Currently `update` does the same thing as `replace`, for details, refer to https://github.com/strongloop/loopback-connector-cloudant#no-partial-update
 
+## Find By Id
+
+See the advanced queries documentation [Find By Id](./doc/find-by-id.md)
+
 # Migration
 
 After attaching a model to a Cloudant datasource, either statically with `model.json` file or dynamically in boot script code, user need to run `automigrate` or `autoupdate` to migrate models to database. Cloudant connector does NOT automatically migrate them.
--- a/doc/find-by-id.md
+++ b/doc/find-by-id.md
@@ -1,0 +1,21 @@
+# Find By Id
+
+When find an instance by its id, you can invoke the connector level method `findById` to use the lookup instead of query under the hood.
+
+*/server/script.js*
+```javascript
+module.exports = function(server) {
+  // Get Cloudant dataSource as `ds`
+  // 'cloudantDB' is the name of Cloudant datasource created in 
+  // 'server/datasources.json' file
+  var ds = server.datasources.cloudantDB;
+
+  ds.once('connected', function() {
+    // 1. Please note `ds.connector.findById()` is the correct way to call it.
+    // 2. This api matches the Cloudant driver's function db.get(),
+    //    instead of db.find()
+    ds.connector.findById(model_name, your_instance_id, function(err, results) {
+      // `results` would be the raw data returned from the driver
+    });
+};
+```


doc/find-by-id.md
```diff
@@ -0,0 +1,21 @@
+# Find By Id
+
+When find an instance by its id, you can invoke the connector level method `findById` to use the lookup instead of query under the hood.
```
bajtos, line 3:
```suggestion
To find an instance by its id, you can invoke the connector level method `findById` to use the lookup instead of query under the hood.
```

Can you please explain what are the ramifications of this approach? I believe the public implementation of `findById` is using a query to ensure that `Product.findById('123')` does not return a category model instance when `'123'` is an id of a category object in the database.

README.md
```diff
@@ -633,6 +633,10 @@ We are still in progress of refactoring some methods, more details to be updated
 
 Currently `update` does the same thing as `replace`, for details, refer to https://github.com/strongloop/loopback-connector-cloudant#no-partial-update
 
+## Find By Id
+
+See the advanced queries documentation [Find By Id](./doc/find-by-id.md)
```
bajtos, line 638: Is it worth creating a new page for those ~20 lines? Is this link going to work when viewed at https://www.npmjs.com/package/loopback-connector-cloudant?

bajtos:
Recently, I was helping a colleague to implement `findById` version that bypasses model name checks to use `get` instead of `query` method. While this is not entirely safe (it makes it possible to fetch object of "Product" type by calling "Category" repository method), it may be ok in some cases.

```ts
export class CloudantRepository<T extends Entity, Relations extends object = {}> extends DefaultCrudRepository<T, string, Relations> {
  async findById(id: string, filter?: FilterExcludingWhere<T>, options?: Options): Promise<T & Relations> {
    const connector = this.dataSource.connector!;
    const cloudant = connector.cloudant;
    const db = cloudant.use(connector.getDbName(connector));

    const result = await db.get(id);
    const persistedModelInst = new this.modelClass(result);
    return this.toEntity<T & Relations>(persistedModelInst);
  }
}
```

Is this perhaps something to capture in our docs too?

doc/find-by-id.md
jannyHou: @bajtos My bad, I forgot this PR :( 

There is a long story of why we chose to use connector level `findById`, the discussion is here: https://github.com/strongloop/loopback-connector-couchdb2/issues/46 and https://github.com/strongloop/loopback-datasource-juggler/pull/1459

tl;dr juggler's `findById` invokes `find` with filter, but doesn't call connector level's `findById` directly, and therefore doesn't have all the complicated processing logic before/after getting the connector's result.

But Cloudant has a much cheaper plan for query by id.

README.md
jannyHou: sure I can move it back to the readme :p

doc/find-by-id.md
```diff
@@ -0,0 +1,21 @@
...
+
+*/server/script.js*
+```javascript
+module.exports = function(server) {
+  // Get Cloudant dataSource as `ds`
+  // 'cloudantDB' is the name of Cloudant datasource created in 
+  // 'server/datasources.json' file
+  var ds = server.datasources.cloudantDB;
```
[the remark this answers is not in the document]
jannyHou: yep it's for lb3 only.

jannyHou:
Thanks @bajtos I will add calling `findById` in lb4 as well. 

For:

> it makes it possible to fetch object of "Product" type by calling "Category" repository method

Are you concerning about 

```ts
const couchModel = connector.selectModel(this.modelClass.modelName);
```

can select any model?


And couchdb2 connector actually exposes the `findById` method(cloudant inherits it), which takes in a model name as 1st parameter, see

https://github.com/strongloop/loopback-connector-couchdb2/blob/e43ee44ef47439b967b796182eef08c81ae8051d/lib/couchdb.js#L491

bajtos:
> it makes it possible to fetch object of "Product" type by calling "Category" repository method

Here is the scenario I have in mind, the code is using LB4 repository API and assuming that `findById` is implemented using one of the methods we discussed earlier, calling `db.get(id)`.

```ts
const category = await categoryRepo.create({name: 'Stationery'});
const customer = await customerRepository.findById(category.id);
// What happens now? 
// Did we create a Customer instance filled with data coming from a Category?
```

bajtos:
IMO, we should use the following user story to frame the goals and focus of the work related to `findById` for Cloudant.

_As a user building LB4 REST API backed by Cloudant database, I want an efficient `findById` implementation (an efficient `GET /{models}/{id}` endpoint) that IS NOT consuming expensive `query` API calls._

Global queries are expensive both cost wise ($5.00 USD/Global Query capacity vs. $0.25 USD/Read capacity) and performance wise (API call limit is 100 reads/sec but only 5 queries/sec) - see https://cloud.ibm.com/docs/Cloudant?topic=Cloudant-ibm-cloud-public

dhmlau:
@jannyHou, are you still interested in pursuing this PR?

jannyHou:
@dhmlau I would like to finish it, will find some time to continue asap

bajtos:
@jannyHou ping :) 🔔

jannyHou:
> Here is the scenario I have in mind, the code is using LB4 repository API and assuming that `findById` is implemented using one of the methods we discussed earlier, calling `db.get(id)`.

```ts
const category = await categoryRepo.create({name: 'Stationery'});
const customer = await customerRepository.findById(category.id);
// What happens now? 
// Did we create a Customer instance filled with data coming from a Category?
```


I think yes, `mo.db.get(id,cb)` only cares about the `id` field, it doesn't check other properties. We must manually verify the fetched instance belong to the query model.

jannyHou:
@bajtos I submitted a PR to promisify the findById method so that lb4 app can call it by `await connector.findById(modelName, idValue)`, see PR https://github.com/strongloop/loopback-connector-couchdb2/pull/83, it also makes sure the fetched instance belong to query model.


Here is the repository method I tested with:

```
  async findById(id: string, filter?: FilterExcludingWhere<Todo>, options?: Options): Promise<Todo & TodoRelations> {
    const connector = this.dataSource.connector!;
    // const couchModel = connector.selectModel(this.modelClass.modelName);
    try {
      // I tried to use `promisify(connector.findById)`, but the invoked method has `this` as `undefined`
      // So I submitted a PR to promisify the connector level method.
      const result = await connector.findById('Todo', id);
      console.log(result);
      return result;
    } catch (error) {
      if (error.message.match(/not found/)) {
        throw new EntityNotFoundError(this.entityClass, id);
      } else {
        throw error;
      }
    }
  }
```

bajtos:
Thank you @jannyHou for the update. Will your repository method correctly handle "not found" case? The connector implementation of `findById` is returning an empty array for some "not found" conditions. Other than that, your code snippet looks great!

dhmlau:
Abandoned.

